PR #1, opened 2024-10-02
Implement vector math prototype

joaander:
Implement vector math via a `Vector` trait.

TODO:
- [x] Implement `fmt::Display` for `CartesianVector<N>`.
- [x] Implement `Neg` for `CartesianVector<N>`. Also add it to the supertraits of `Vector`.
- [x] Implement `Cross` for `CartesianVector<3>`.
- [x] Unit test all methods implemented for `CartesianVector<N>`.
    - [x] Sub/Mul/Div and assign variants
    - [x] Cross
    - [x] CartesianVector::from/try_from
- [x] Benchmark additional methods.
    - [x] From array/from vec 
    - [x] Sub/Mul/Div
- [x] Remove `vec3.rs` - assuming that `CartesianVector<3>` is just as fast.
    - [x] Implement `From` methods to compare efficiency
- [x] Write API documentation.

For a future PR: 
- Implement quaternions (and rotations by angles in 2D? -> yes please)


janbridley:
I wrote `Display` for the cartesian vector using what appears to be a "standard" approach, resulting in a string in the form "[n, n, n, ... n]". ~I will continue iterating (for learning's sake) to see if there are simpler implementations~. While I found a few alternatives - notably the one below - they are less readable in my opinion. 

```rust
impl<const N: usize> fmt::Display for CartesianVector<N> {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        write!(
            f,
            "[{}]",
            self.coordinates
                .iter()
                .map(|x| x.to_string())
                .reduce(|a, b| a + "," + &b)
                .unwrap_or_default()
        )
    }
}
```

We could add line wrapping for vectors with N > (80-2)/3=26, but this seems unnecessary

joaander:
Nice! I don't have a strong preference for reduce vs join for display.

By the way, feel free to make use of rstest as we discussed to make it easier to test multiple cases. All the tests I wrote were just a scratch pad where I was figuring out how the Add (and other) operators function.

janbridley:
Creating a declarative (`macro_rules!`) macro that parameterizes tests with const generics is relatively easy, but it appears that a procedural-attribute macro (`#macro(param)`) that handles constants is extremely difficult. Proc macros also require another crate (and likely the large Syn library, if we want clean handling of edge cases). I plan to move forward with the declarative macro implemented in the last commit for parameterizing const generics as an initial prototype

hoomd-rs-vector/src/cartesian.rs
```diff
@@ -26,18 +26,42 @@ impl<const N: usize> From<[f64; N]> for CartesianVector<N> {
     }
 }
 
+impl<const N: usize> From<Vec<f64>> for CartesianVector<N> {
```
joaander: `From` should not be fallible: https://stackoverflow.com/questions/73928188/is-it-idiomatic-to-use-the-from-into-trait-for-fallible-conversions

We should implement:
* `TryFrom<Vec<f64>>`
* `From<std::ops::Range<usize>>`
* `FromIterator<f64>`
so that the caller can handle the failure.

hoomd-rs-vector/src/cartesian.rs
```diff
@@ -26,18 +26,42 @@ impl<const N: usize> From<[f64; N]> for CartesianVector<N> {
...
+    #[inline]
+    fn from(coordinates: Vec<f64>) -> Self {
+        let coordinates = coordinates
+            .try_into()
+            .expect("Failed to initialize CartesianVector from Vec<f64>");
+        Self { coordinates }
+    }
+}
+
 impl<const N: usize> From<std::ops::Range<usize>> for CartesianVector<N> {
```
joaander: The conversions from `std::ops::Range<usize>` and `FromIterator<f64>` make a temporary dynamic allocation (when they collect into `Vec`). We should note this in the documentation so that callers are aware of the possible performance implications.

janbridley: I found a way to get around this dynamic allocation, with approximately the same level of complexity: see [here](https://github.com/glotzerlab/hoomd-rs/pull/1/commits/420be9ae71c5562e7ad1f6baf84d7b8af566b082)

hoomd-rs-vector/src/cartesian.rs
janbridley: This has definitely been a learning experience - I found a clean way to swap to `TryFrom<Vec<f64>>`. It makes sense that the `From<std::ops::Range<usize>` is infallible, as the range has a defined length. 

I have a few conceptual questions about `FromIterator`, as there is no `TryFromIterator` (afaik) and the array building could fail. This may be a good in-person discussion for the working meeting, as I'm curious how to keep things "idiomatic" here

janbridley:
I added benchmarks for `from<[f64; N]>` and `try_from<Vec<f64>>`. I can't tell if the benchmark is being compiled out for the array conversion, as the benchmark takes on the order of picoseconds which seems unreasonable. The try_from<vec> takes about as long as I would expect.

I established a slightly different pattern for these tests, as I could not find a way to iterate over a pre-generated vec of data without cloning the elements in cases (which would skew the performance results). The new pattern allows `divan` to auto select the number of samples, and generates a new piece of test data for each benchmark loop. This is a sidegrade rather than a strict improvement, but it does seem to follow divan's user guides. I used the `rand` crate to generate random numbers, but we should consider establishing a range/distribution for tests moving forward

hoomd-rs-vector/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+// nursery
+#![warn(
+    clippy::fallible_impl_from,
+    clippy::needless_collect,
+    clippy::needless_pass_by_ref_mut
+)]
+#![deny(clippy::expect_used, clippy::panic, clippy::unwrap_used)]
+#![warn(missing_docs)]
+
+//! Vector and quaternion math.
```
janbridley: Should we use [block comments](https://doc.rust-lang.org/reference/comments.html#doc-comments) for this?

joaander: Indeed we should, block comments are a lot easier to add content to (no need to copy `//!` to the start of each line).

hoomd-rs-vector/ARCHITECTURE.md
```diff
@@ -0,0 +1,44 @@
...
+## Vector
+
+The `hoomd_rs_vector` crate defines a generic `Vector` trait that is independent of
+representation. The trait consists of methods that can _only_ be applied to all vectors
+in a vector space with *n* dimensions:
+
+- Vector addition/subtraction
+- Multiplication by a scalar
+- Dot product
+- Length / length squared
```
janbridley:
```suggestion
- Vector addition & subtraction
- Multiplication by a scalar
- Dot product
- Length & length squared
```

janbridley:
Should we implement a `from_random` method (or similar) for the cartesian vectors? It's a pattern we use frequently, so it may be convenient

hoomd-rs-vector/src/cartesian.rs
```diff
@@ -0,0 +1,566 @@
...
+        ))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use paste::paste;
+
+    macro_rules! parameterize_vector_length {
```
janbridley:
```suggestion

    // Parameterize a test function over an array of vector lengths
    macro_rules! parameterize_vector_length {
```

hoomd-rs-vector/src/lib.rs
```diff
@@ -0,0 +1,342 @@
...
+    /// Attempted converting a value to a vector with a dimension not equal to the value's length.
+    #[error("Source does not match the target vector length.")]
+    InvalidVectorLength,
+}
+
+/** Placeholder type for use in trait bounds on [`Vector`] dimension.
+
+    This placeholder will be removed when Rust allows the use of associated constants in trait
+    bounds.
+*/
```
janbridley: It would be great to link a PR or RFC here for tracking - is it the same as below (https://github.com/rust-lang/rust/issues/92827)?

joaander:
> Should we implement a `from_random` method (or similar) for the cartesian vectors? It's a pattern we use frequently, so it may be convenient

Yes. The `rand` crate way of doing this is to implement the `Standard` distribution and/or other distributions as needed for custom types: https://rust-random.github.io/book/guide-dist.html#uniform-sampling-by-type Then you can sample vectors from the distribution:
```
let v: CartesianVector<3> = rng.gen()
```
or
```
let ball = hoomd_rs_vector::distributions::Ball(5.0)
let v: CartesianVector<3> = rng.sample(&ball)
```

I think the `Standard` distribution for rotations should be a uniformly randomly selected rotation. For vector, I'm not sure if it should be uniform in the unit cube or in the unit sphere. In either case, we will want distributions for random vectors in a ball, on a sphere (possibly) and random "small" rotations (for MC trial moves).

hoomd-rs-vector/src/lib.rs
joaander: Yes, that is the correct tracking issue to link.

joaander:
Merging now. We can refactor further in a future PR if needed.

PR #2, opened 2024-10-27
Add random vector implementation

janbridley:
`rand` makes it [extremely easy](https://docs.rs/rand/latest/rand/distributions/struct.Standard.html#custom-implementations) to implement random generators for custom types, and by enabling the "min_const_gen" flag for the package we can trivially generate arrays of any size. This implementation lets `Cartesian` be generated from any rand::distribution, with the default  `Standard` distribution being a unit hypercube.

Sidebar - it would be nice to increase the struct_lit_width rustfmt setting to allow slightly longer lines initializing a struct. For example:
```rust
//Before
Self {
	coordinates: coordinates.into()
}

//After
Self { coordinates: coordinates.into() }
```

```toml
struct_lit_width = 32 # Default is 20, anywhere 32-48ish works well imo
```

### DIFF
--- a/hoomd-rs-vector/Cargo.toml
+++ b/hoomd-rs-vector/Cargo.toml
@@ -13,11 +13,11 @@ version.workspace = true
 
 [dependencies]
 thiserror = "1.0.64"
+rand = {version = "0.8.5", features=["min_const_gen"]}
 
 [dev-dependencies]
 divan = "0.1.14"
 paste = "1.0.15"
-rand = "0.8.5"
 
 [[bench]]
 name = "cartesian"
--- a/hoomd-rs-vector/Cargo.toml
+++ b/hoomd-rs-vector/Cargo.toml
@@ -13,7 +13,7 @@ version.workspace = true
 
 [dependencies]
 thiserror = "1.0.64"
-rand = {version = "0.8.5", features=["min_const_gen"]}
+rand = {version = "0.8.5"}
 
 [dev-dependencies]
 divan = "0.1.14"
--- a/hoomd-rs-vector/src/vector.rs
+++ b/hoomd-rs-vector/src/vector.rs
@@ -9,8 +9,8 @@ use std::fmt;
 use std::iter::zip;
 use std::ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Neg, Sub, SubAssign};
 
-use rand::Rng;
 use rand::distributions::{Distribution, Standard};
+use rand::Rng;
 
 use crate::{Cross, Dimension, Error, Vector};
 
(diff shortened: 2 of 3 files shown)


hoomd-rs-vector/src/vector.rs
```diff
@@ -297,6 +300,29 @@ impl Cross for Cartesian<3> {
...
+impl<const N: usize> Distribution<Cartesian<N>> for Standard {
+    /** Create a Cartesian vector with coordinates drawn from a unit (hyper)cube.
+
+    ```
+    # use hoomd_rs_vector::vector;
+    # use rand::{thread_rng, Rng};
+    # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let mut rng = rand::thread_rng();
+    let v: vector::Cartesian::<3> = rng.gen();
+    // assert_eq!(v, [3.0, 4.0, 5.0].into());
```
janbridley: TODO: it would be nice to have a simple test or verification line here in place of the assertion - I'm open to suggestions

hoomd-rs-vector/src/vector.rs
```diff
@@ -564,4 +590,19 @@ mod tests {
         );
     }
     parameterize_vector_length!(from_vec, [2, 3, 4, 8, 16, 32]);
+
+    fn random_in_range<const N: usize>() {
+        // At a minimum, this check should verify that we are drawing from a unit cube
```
janbridley:
```suggestion
        // Loosely verify we are drawing from the correct distribution
```

janbridley:
While I think a final version could include an image in the docstring showing the distribution range, this should be ready to merge!

hoomd-rs-vector/Cargo.toml
```diff
@@ -13,11 +13,11 @@ version.workspace = true
 
 [dependencies]
 thiserror = "1.0.64"
+rand = {version = "0.8.5", features=["min_const_gen"]}
```
joaander: Do we still need the `min_const_gen` feature?

joaander (commented):
Thanks! Images are currently only possible in rustdoc if we host the image on a separate webserver. There is a request from 2016 (https://github.com/rust-lang/rust/issues/32104), but it seems unlikely to be implemented any time soon.

hoomd-rs-vector/Cargo.toml
joaander: It is not. I removed it.

joaander (approved):
Thanks! This is very clean code and provides a clear template for future distributions.

hoomd-rs-vector/Cargo.toml
janbridley: Thank you!

PR #3, opened 2024-11-07
Add rotation/quaternion infrastucture

DomFijan:
This PR implements rotation algorithms and traits for 2D (angle) and 3D (quaternion) rotation and general rotation using rotation matrices.


janbridley:
I added the Index trait for vectors, which allow for some nice syntax. I will go through and update methods to use the syntax in the morning if it seems like something that would be useful.

joaander:
> I added the Index trait for vectors, which allow for some nice syntax. I will go through and update methods to use the syntax in the morning if it seems like something that would be useful.

Thanks! For `Angle`, I added `Display`, `rand::Standard`, benchmarks, a `normalized` method, and precomputed rotations. I also modified the examples to consistently show `use` statements and be placed under an `## Example` heading.

I think this is ready for a quick review so it can be merged soon. We can iteratively improve it as we start to put it to use. I'll give Quaternion a similar treatment tomorrow.

hoomd-rs-vector/src/rotation/angle.rs
```diff
@@ -0,0 +1,396 @@
...
+pub struct Angle {
+    /// Rotation angle (radians).
+    pub theta: f64,
+}
+
+/// A precomputed rotation about an [`Angle`]
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Precomputed {
+    /// Rows of the rotation matrix.
+    row: [Cartesian<2>; 2],
```
janbridley: I find `row` to be a bit confusing - perhaps `rows` is better? The line in `impl Rotate for Precomputed ... self.row[0].dot(vector)` is what threw me off initially, although this is absolutely a nitpick so we can go ahead regardless

janbridley (approved):

hoomd-rs-vector/src/rotation/angle.rs
joaander: done

PR #5, opened 2025-01-17
Interaction

joaander:
Implement `hoomd_interaction` - the low level functions that define the energy and force interactions between particles.

All methods implemented so far are pairwise interactions in the `pairwise` module. I expect to add other modules in the future (e.g. angle, dihedral, external, ...)

### DIFF
--- a/hoomd-interaction/benches/angular-mask.rs
+++ b/hoomd-interaction/benches/angular-mask.rs
@@ -12,9 +12,9 @@ use divan::{self, black_box, Bencher};
 use rand::{rngs::StdRng, Rng, SeedableRng};
 use std::f64::consts::PI;
 
-use hoomd_interaction::pairwise::{AnisotropicEnergy, AngularMask, LennardJones};
 use hoomd_interaction::pairwise::angular_mask::Patch;
-use hoomd_vector::{Cartesian, Angle, Versor};
+use hoomd_interaction::pairwise::{AngularMask, AnisotropicEnergy, LennardJones};
+use hoomd_vector::{Angle, Cartesian, Versor};
 
 fn main() {
     divan::main();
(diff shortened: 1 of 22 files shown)


janbridley:
XPLOR has been added, docs are still TODO and a few tests could be expanded upon

hoomd-interaction/src/pairwise/xplor.rs
```diff
@@ -0,0 +1,152 @@
...
+            let r_cut_sq = self.r_cut.powi(2);
+            let r_on_sq = self.r_on.powi(2);
+            (r_cut_sq - r_sq).powi(2) * (r_cut_sq + 2.0 * r_sq - 3.0 * r_on_sq)
+                / (r_cut_sq - r_on_sq).powi(3)
+        }
+    }
+    /// The xplor shifting function
+    #[inline]
+    fn ds_dr(&self, r: f64) -> f64 {
+        // TODO: can we share r_*sq between energy and force?
```
joaander: I thought about this. If we wanted to be sure, we could implement a separate `EnergyAndForce` trait. However, this makes multiple ways to compute the same thing - which is confusing to users and may lead to inconsistencies or bugs.

We are specifying `#[inline]` everywhere. The optimizer should be smart enough to remove the duplicate calculations when there is code that calls both `.force` and `.energy` in the same function.

janbridley: Sounds good - I'm inclined to let the compiler figure it out as well. I'll take a peek with godbolt if I have the time later today

janbridley: ~Godbolt did not seem to honor my `[#inline]` request but I don't have much experience using it. For now I think the simplicity of the current implementation is fine~ This simple example appears to work as expected:

```rust
use std::hint::black_box;
#[inline]
pub fn square(num: f64) -> f64 {
let num_sq =  num.powi(3);
num_sq
    
}
#[inline]
pub fn quat(num: f64) -> f64{
let num_sq = num.powi(3);
num_sq * num_sq
}

#[inline(never)]
pub fn run(x: f64)->f64 {
 square(x) + quat(x)
}
pub fn main() { 
  let x = 96.5132;
  black_box(run(x));
}

```

```asm
# Compilation provided by Compiler Explorer at https://godbolt.org/
example::run:
        movapd  xmm1, xmm0
        mulsd   xmm1, xmm0
        mulsd   xmm0, xmm1
        movapd  xmm1, xmm0
        mulsd   xmm1, xmm0
        addsd   xmm0, xmm1
        ret

.LCPI1_0:
        .quad   0x405820d844d013a9
example::main:
        push    rax
        movsd   xmm0, qword ptr [rip + .LCPI1_0]
        call    qword ptr [rip + example::run@GOTPCREL]
        movsd   qword ptr [rsp], xmm0
        mov     rax, rsp
        pop     rax
        ret


# Execution result with exit code 0
```

hoomd-interaction/benches/angular-mask.rs
```diff
@@ -0,0 +1,73 @@
...
+        .counter(ItemsCount::from(1_u32))
+        .with_inputs(|| -> (Cartesian<3>, Versor) { (rng.gen::<Cartesian<3>>(), rng.gen::<Versor>()) })
+        .bench_local_values(|(r_ij, angle)| black_box(angular_mask.energy(&r_ij, &angle)));
+}
+
+// To inspect the energy function with cargo-show-asm
+// #[no_mangle]
+// pub fn asm(angular_mask: &AngularMask<LennardJones, Cartesian<3>>, r_ij: &Cartesian<3>, v: &Versor) -> f64 {
+//     angular_mask.energy(r_ij, v)
+// }
```
joaander, line 101: @janbridley FYI, The commented code is how examined the asm for more complex examples. I didn't want to try and copy and paste all of hoomd-vector into the web page service.

Using the `cargo-show-asm` package, the following command showed me the asm I was expecting (with the above code uncommented, of course):
```
cargo asm -p hoomd-interaction --bench angular-mask "<hoomd_interaction::pairwise::angular_mask::AngularMask<F,V> as hoomd_interaction::pairwise::AnisotropicEnergy<V,R>>::energy" --rust
```
I'm not sure why:
```
cargo asm -p hoomd-interaction --bench angular-mask "asm"
```
didn't work (it showed that `asm` called the above fn). Maybe it would if I had followed cargo-show-asm's advice and marked the `asm` fn `#[inline(never)]` instead of `#[no_mangle]` - but the 2nd option is a hint I found somewhere on stackoverflow.

PR #6, opened 2025-01-23
Shapes

janbridley:
## Overview

This PR adds a wide variety of geometry-related functionality to HOOMD. The basic trait structure for geometries has been implemented, as have a small set of commonly-used geometric primitives. Code for intersections has been added and (extensively) tested, with some discussion below.

Note that, in development of the ellipse fast overlap check, some simple matrix functionality became useful - I scratched some thoughts down in `hoomd_geometry/shapes/ellipsoid.rs`, but these should be re-evaluated and thought out more carefully. @tfteague expressed interest in working on a `hoomd_matrix` or similar, but no rush either way. For now, I can remove these methods and comment out the fast ellipsoid overlap, as there's enough docs left to add to `hoomd_geometry` to focus on for now. Note that, as expected, the ellipsoid overlap highlighted some failures in xenocollide for rounded shapes - I'll look into these in a separate PR, but we are at least able to test the precision now.

Also if anyone wants to look over the overlap check method, feel free! After trying and failing to implement a few methods from various papers, I implemented my own solution that appears to work fairly well (and agrees with HOOMD or a visual inspection of the intersection for > 1e5 samples). Although it is an iterative solution, in most cases it converges within a single iteration as we only need to find a single negative point on a function. I could probably use Sturm sequences to get even faster, but I'm happy with the performance for now.

#### TODOs: 

- [x] Xenocollide precision checks.
    - [x] 2d
    - [x] 3d
- [x] Translate this PR comment to architecture.md
- [x] ~Add in particle radius (bound on Shape trait) for Xenocollide~ Should be a separate PR
- [x] ~Tune tolerances for f64: Hoomd-Blue also fails one test with HOOMD_SHORTREAL_PRECISION=32, so we are at least in "expected behavior" territory~ tuning the tolerance will not fix edge cases where a dot product ends up as less than zero, so this tuning can be saved fora later date where we have a real, failing test case
- [x] Tetrahedron fast overlap check (for validation)
- [x] **Ellipse fast overlap check (for validation with rounding)**!
- [x] Use `Option<Rotation>` for rotation in IntersectsAt trait, as None rotation implies no change and allows for more accurate switching between overlap check modes.
- [x] Rename `r_ij` -> `v_ij`
- [x] Update `collide2|3d` to take a `point_deep_inside` parameter that ensures the model's correctness across varied inputs. This change does not impact primitives with alternate intersection methods

### Traits
The `Shape` trait provides widely useful and easy to implement functionality for geometric primitives. The methods of this trait have been purposefully left "simple" (e.g. `bounding_sphere`, rather than `minimal_bounding_sphere`) in order to broaden the number of structs on which it can be implemented. More specific (and possibly more complex functionality) is included in additional traits. `Volume` is a good example, as it defines the n-hypervolume of the primitive which may or may not be easy to calculate. 

More specific functionality is included in additional traits: `SupportFn` defines the support function of a geometry, and is used with many implementations of the `IntersectsAt` trait to determine the intersection of two bodies.

In general, primitives provided by this crate _do not_ include a notion of the centroid or orientation of the primitive. While in some cases (particularly polytopes) do intrinsically carry this information, in general we assume that that primitives are centered at the origin and rotated by the identity matrix of the correct dimension. This definition allows for greater portability between this crate and more simulation-oriented tools, which require an external definition of position and orientation for performance reasons.
#### Intersections

The `IntersectsAt` trait makes up a significant portion of the code in this crate, and provides a wide variety of methods that allow for the calculation of overlaps between geometric primitives. Most provided methods are an overlap _test_, taking in two geometries, a displacement `Vector`, and a `Rotation` and returning a boolean indicating whether the geometries intersect. Many shapes -- including spheres, oriented cuboids, and tetrahedra -- implement shape-specific overlap checks that are faster than general methods for determining overlaps. An implementation of the Xenocollide collision detection test is included in the `collide[2|3]d` functions, and will function properly for any convex geometry that implements `SupportFn<V: Cartesian<[2|3]>>`. While all methods should provide the same results, more specialized subroutines often provide greater performance than Xenocollide.

~The `IntersectsAt` trait includes a helper type that allows implementations to accept either an `&R: Rotation` or `&Option<R: Rotation>`. This allows for much clearer distinction of axis-aligned intersection modes for AABBs, and ensures users do not have to initialize an `Angle` or `Versor` for sphere overlaps.~ Addressed with Convex wrapper

~Note that, although the following code is valid, such an implementation precludes specific, optimized overlap methods. Instead, `T` should be replicated with a specific type.~ Addressed with Convex wrapper
```rust
impl<S: SupportFn<Cartesian<3>>, R: Rotate<Cartesian<3>> + Rotation + Copy, T>
    IntersectsAt<S, Cartesian<3>, R> for T
where
    RotationMatrix<3>: From<R>,
    T: SupportFn<Cartesian<3>>,
{
    /// Determine whether a convex object intersects another shape at some position and orientation.
    type OptionalRotation = R;
    #[inline]
    fn intersects_at(&self, other: &S, v_ij: &Cartesian<3>, o_ij: &R) -> bool {
        collide3d(self, other, v_ij, o_ij)
    }
}
```

To implement `IntersectsAt` for concave geometries, subdivide the primitive into convex subsets and apply the appropriate collision detection algorithms. While no examples are currently included with `hoomd-geom` a `ShapeUnion` struct would be the most natural extension.

#### Dimension-agnostic primitives

- `Sphere<N>`
- `Cuboid<N>`: axis-aligned boxes
- `ConvexPolytope<N>`
- `HyperEllipsoid<N>`

#### Dimension-specific primitives
- `Simplex3`: Tetrahedron
- `Capsule`: Spherocylinder

#### Encapsulating structs

- `Centered`
- `Sphero`

## Discussion

This crate is mostly complete, but there are a few outstanding bugs with the overlap code. 

- [x] collide[2|3]d fails for spheres when a non-zero rotation is passed. Spheres are not orientable, but the code should work regardless of the input
- [x] collide[2|3]d fails some checks that HOOMD's implementation passes. The python file in the main directory enumerates these examples as well a HOOMD test script 
- [x] Simplex3::intersects_at occasionally detects an overlap when near certain types of intersections


hoomd-geom/src/intersects.rs
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) 2024-2025 The Regents of the University of Michigan.
+// Part of hoomd-rs, released under the BSD 3-Clause License.
+
+use crate::Sphere;
+use hoomd_vector::{Cartesian, Vector};
+// use crate::Shape; // TODO: do we want this as a trait bound on S?
+
+pub trait Intersects<S> {
+    fn intersects(&self, other: S) -> bool;
```
joaander: How does one check for intersections between two shapes at different positions and orientations? Shouldn't this have an `r_ij` and `o_ij` argument like I implemented in `AnisotropicEnergy`? Or do you envision a pure `Intersects` for base shapes and a different trait for intersections between shapes that have a position and orientation?

janbridley: This is a good question - my initial thought was that the base `Intersects` trait would be purely geometric, allowing a single trait to handle oriented and non-oriented shapes. However, I'm now realizing two problems with this: it requires shapes to store their own position data (which makes sense for spheres but less sense for other geometries), and it requires the creation of a new shape when applying orientations. This second condition is probably fine for copyable types, but definitely is not for non-copy types.


I think this raises a more general question about geometry code for simulations vs. other applications. A "sphere" should almost certainly contain some information about its position for geometric calculations, but for a simulation this is redundant. I am beginning to think the solution to this is differentiating between a `Particle` (minimal information needed to specify a shape, given that position and orientation data is provided elsewhere) and a `Shape` (minimal information needed to fully define a shape). Let me know your thoughts!

janbridley:
There is still some work to be done (primarily tests), but this code is now in a much better place. I have complete `intersects_at` implementations for Sphere and Cuboid, with an example for how a method could enforce the provided rotation to be the identity. I'd love another set of eyes on this to let me know if this progress looks to be in the right direction!

hoomd-geom/src/xenocollide.rs
```diff
@@ -204,10 +204,83 @@ pub fn collide3d<R: Rotate<Cartesian<3>> + Rotation + Copy, T: SupportFn<Cartesi
...
+        } else {
+            if v_perp_v4v0.dot(&v3) > 0.0 {
+                v2 = v4; // OUTside v1 && inside v3  => eliminate v2
+            } else {
+                v1 = v4; // OUTside v1 && OUTside v3 => eliminate v1
+            }
+        }
 
-    // TODO: continue
+        /* Match case is way cleaner but less efficient (calcualtes all 3 dot products)
```
joaander: Have you measured that it is less efficient? One could hope that the compiler is able to defer the computation of the dot product until the particular match arm is evaluated.

janbridley: It appears it's performance equivalent but I want to make sure given that this is the hottest loop in HPMC. I'd love to chat about Cargo ASM at some point to verify

hoomd-vector/src/cartesian.rs
```diff
@@ -407,6 +446,24 @@ where
...
+impl Cartesian<2> {
+    /// Compute a 2-vector perpendicular to self.
+    #[inline]
+    #[must_use]
+    pub fn perp(self) -> Self {
+        Cartesian::from([-self[1], self[0]])
+    }
+}
+
+impl<const N: usize> Cartesian<N> {
```
janbridley: TODO: implement at the Vector level

hoomd-vector/src/cartesian.rs
```diff
@@ -274,20 +317,16 @@ impl<const N: usize> Mul<f64> for Cartesian<N> {
...
-        for (result, a) in coordinates.iter_mut().zip(self.coordinates) {
-            *result = a * rhs;
+        Self {
+            coordinates: self.coordinates.map(|x| x * rhs),
         }
-        Self { coordinates }
     }
 }
 
 impl<const N: usize> MulAssign<f64> for Cartesian<N> {
```
janbridley, line 335: TODO: test out a "broadcast" method and make sure it compiles to the same code

janbridley:
@joaander While looking at adding a point_deep_inside parameter to xenocollide, I realized that our construction of intersection methods using a delta vector `v_ij` makes this information redundant. Assuming a point `p` deep within `B⊖A` is found by `point_deep_inside_b-point_deep_inside_a`, then our expression for `p` is `(point_deep_inside_a+v_ij)-point_deep_inside_a` simplifies to just `v_ij` regardless of whether the origin is contained within the shape. Did I miss a step here, or does this seem correct?

Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [workspace]
 
 members = [
+  "hoomd-geom",
```
joaander: Can we name it `hoomd-geometry`? Full words make the code easier to understand, especially for non-native speakers. When I read this, my eyes skip the `m` and I see [Geo](https://hollowknight.fandom.com/wiki/Geo).

hoomd-geom/ARCHITECTURE.md
```diff
@@ -0,0 +1,49 @@
...
+
+The `IntersectsAt` trait includes a helper type that allows implementations to accept either an `&R: Rotation` or `&Option<R: Rotation>`. This allows for much clearer distinction of axis-aligned intersection modes for AABBs, and ensures users do not have to initialize an `Angle` or `Versor` for sphere overlaps.
```
joaander: Update this to reflect the current code. We opted for a different solution.

hoomd-geom/ARCHITECTURE.md
```diff
@@ -0,0 +1,49 @@
...
+
+Note that, although the following code is valid, such an implementation precludes specific, optimized overlap methods. Instead, this method should be implemented for
```
joaander: Maybe instead of noting this, maybe outline the specifically optimized (non-generic) versions?
```
impl SupportFn<Cartesian<3>> for CustomShape {
...
}

impl IntersectsAt<CustomShape, Versor> for T {
...
}
```

hoomd-geom/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+rstest = "0.24.0"
+
+[lints]
+workspace = true
+
+[dev-dependencies]
+approx = "0.5.1"
+divan = "0.1.16"
+paste = "1.0.15"
+rand = "0.9.1"
```
joaander: After merging the latest `trunk` into the branch, you can replace this with:
```suggestion
approx.workspace = true
divan.workspace = true
rand.workspace = true
rstest.workspace = true
```
to inherit the versions from the workspace.

I removed `paste` from the list because it doesn't appear to be used.

hoomd-geom/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+license.workspace = true
+keywords.workspace = true
+categories.workspace = true
+description = "Performant, general computational geometry methods in Rust."
+readme = "hoomd-geom/README.md"
+
+[dependencies]
+hoomd-vector = { path= "../hoomd-vector" }
+itertools = "0.14.0"
+rstest = "0.24.0"
```
joaander:
```suggestion
```
Shouldn't this be a dev dependency?

hoomd-geom/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+repository.workspace = true
+license.workspace = true
+keywords.workspace = true
+categories.workspace = true
+description = "Performant, general computational geometry methods in Rust."
+readme = "hoomd-geom/README.md"
+
+[dependencies]
+hoomd-vector = { path= "../hoomd-vector" }
+itertools = "0.14.0"
```
joaander: `itertools` is only use in tests, move to `dev-dependencies`

hoomd-geom/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+homepage.workspace = true
+repository.workspace = true
+license.workspace = true
+keywords.workspace = true
+categories.workspace = true
+description = "Performant, general computational geometry methods in Rust."
+readme = "hoomd-geom/README.md"
+
+[dependencies]
+hoomd-vector = { path= "../hoomd-vector" }
```
joaander: After merging in trunk, you can do this:
```suggestion
hoomd-vector.workspace = true
```

hoomd-geom/README.md
```diff
@@ -0,0 +1,5 @@
+# hoomd-geom
+
+Computational geometry library for [hoomd].
```
joaander:
```suggestion
Computational geometry library for [hoomd-rs].
```

hoomd-geom/src/common.rs
```diff
@@ -0,0 +1,169 @@
...
+            v_tip / n.norm() + rshift
+        } else {
+            v_base / n.norm() + rshift
+        }
+    }
+}
+
+/// An N-Dimensional [`HyperEllipsoid`] defined by its semi-major axes.
+#[derive(Clone, Copy, Debug)]
+pub struct HyperEllipsoid<const N: usize> {
```
joaander:
```suggestion
pub struct Hyperellipsoid<const N: usize> {
```
Hyperellipsoid is one word. 

Also, if we are using the terminology "hyperellispoid", why not "hypersphere"? Or is that not common in the mathematical literature.

hoomd-geom/src/common.rs
```diff
@@ -0,0 +1,169 @@
+// Copyright (c) 2024-2025 The Regents of the University of Michigan.
+// Part of hoomd-rs, released under the BSD 3-Clause License.
+
+/*! Common geometric primitives that implement only a small number of operations.*/
+use crate::{IntersectsAt, SupportFn, xenocollide::collide3d};
+
+use hoomd_vector::{Cartesian, Rotate, Rotation, RotationMatrix, Vector};
+
+/// A [`Cylinder`] in three dimensions.
+#[expect(dead_code, reason = "End users will make use of this struct.")]
```
joaander: Is this needed now because we write no implementations involving Cylinder? Maybe we should just remove it for now until we decide to implement it.

hoomd-geom/src/common.rs
```diff
@@ -0,0 +1,169 @@
...
+#[derive(Clone, Copy, Debug)]
+pub struct Cylinder {
+    /// Radius of the [`Cylinder`]
+    r: f64,
+    /// Height of the [`Cylinder`]
+    h: f64,
+}
+/// A [`Capsule`] in three dimensions.
```
joaander: Rust doc formats the first line of the doc comment like this:
```
TypeName1     First line
TypeName2     Another first line
```

I find repeating the type name in the first line redundant. I think we should treat the first line as a definition and follow the rule that you can't use a word (type name) to define itself.

hoomd-geom/src/common.rs
```diff
@@ -0,0 +1,169 @@
...
+#[derive(Clone, Copy, Debug)]
+// pub struct Capsule<const N: usize> {
+pub struct Capsule {
+    /// Radius of the [`Capsule`]'s spherical caps.
+    r: f64,
+    /// Distance between the centers of the spherical caps.
+    h: f64,
+}
+
+impl From<(f64, f64)> for Capsule {
```
joaander: I don't think we should be using `From` to initialize shapes, except in very specific cases (i.e. `Hyperellipsoid::From<Sphere>` or `Polytype::From<Cuboid>`).

The Rust docs (https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from) state:
> The conversion is value-preserving: ... String: From<char> is available because they’re both text, but String: From<u32> is not available, since 1 (a number) and "1" (text) are too different.

and

> The conversion is obvious: it’s the only reasonable conversion between the two types.

Without looking at the implementation, I have no idea what (1.0, 2.0) means as a `Capsule`. Same for `Sphere` and the other types. I don't think of a number, i.e. 3.0, as a sphere with that radius. It could just as easily be the diameter.

Unless we come up with good reasons to cache computed values for shapes (and use getter/setter APIs), I think we should make all fields fully public and explicitly named and avoid ambiguous `new` methods (I'm doing that for `hoomd-interaction` in #26). That way, users can initialize with:
```
let sphere = Sphere { radius: 3.0, };
```
and modify with:
```
sphere.radius = 4.0
```
without us having to write any code.

If we want to provide ad-hoc conversion methods for convenience, or need to because we have private internal fields, then we should follow the naming conventions in the Rust API guidelines:
```
let sphere = Sphere::with_radius(3.0);
let sphere2 = Sphere::with_diameter(1.5);
```

hoomd-geom/src/common.rs
```diff
@@ -0,0 +1,169 @@
...
+            v_base / n.norm() + rshift
+        }
+    }
+}
+
+/// An N-Dimensional [`HyperEllipsoid`] defined by its semi-major axes.
+#[derive(Clone, Copy, Debug)]
+pub struct HyperEllipsoid<const N: usize> {
+    /// The principle semi-axes of the [`HyperEllipsoid`] along each direction.
+    axes: Cartesian<N>,
```
joaander: Following the comments above, this should be public.

Also, I don't think of an item in a vector space as defining an ellipsoid. This should be an array:
```suggestion
    axes: [f64; N],
```
If you use methods of `Cartesian` for tricks in the implementation, then you can convert to that representation when needed.

hoomd-geom/src/common.rs
```diff
@@ -0,0 +1,169 @@
...
+    r: f64,
+    /// Height of the [`Cylinder`]
+    h: f64,
+}
+/// A [`Capsule`] in three dimensions.
+#[derive(Clone, Copy, Debug)]
+// pub struct Capsule<const N: usize> {
+pub struct Capsule {
+    /// Radius of the [`Capsule`]'s spherical caps.
+    r: f64,
```
joaander: You have a capsule is defined only by its radius and height. Minimal representations are good, but we need a clear way to define the axis of the capsule. In N dimensions, is the along the first or last? Where is the origin located?

hoomd-geom/src/common.rs
```diff
@@ -0,0 +1,169 @@
...
+    axes: Cartesian<N>,
+}
+impl<const N: usize> IntoIterator for HyperEllipsoid<N> {
+    type Item = f64;
+    type IntoIter = <[f64; N] as IntoIterator>::IntoIter;
+    #[inline]
+    fn into_iter(self) -> Self::IntoIter {
+        self.axes.into_iter()
+    }
+}
```
joaander:
```suggestion
```
I don't think of hyperellipsoids as iterable objects. The axes should be accessible and callers can iterate over the axes field when needed.

hoomd-geom/src/common.rs
```diff
@@ -0,0 +1,169 @@
...
+impl HyperEllipsoid<3> {
+    #[inline]
+    #[must_use]
+    /// Compute a matrix representation of the ellipsoid.
+    #[expect(
+        clippy::many_single_char_names,
+        dead_code,
+        reason = "Ported from HOOMD-Blue, with variable names maintained for consistency."
+    )]
+    fn compute_ellipsoid_matrix<R>(&self, r_ij: &Cartesian<3>, o_ij: &R) -> Cartesian<10>
```
joaander: We should benchmark this in HOOMD-blue and see if it is really any faster than XenoCollide. If the performance increase is marginal or less, I think we can safely remove this code and rely on Xenocollide.

hoomd-geom/src/cuboid.rs
```diff
@@ -0,0 +1,207 @@
...
+use crate::xenocollide::{collide2d, collide3d};
+use crate::{SupportFn, Volume};
+use hoomd_vector::{Cartesian, RotationMatrix};
+use hoomd_vector::{Rotate, Rotation};
+use itertools::multizip;
+use std::cmp::PartialEq;
+
+/** An axis-aligned N-cuboid
+*/
+#[derive(Clone, Copy, Debug)]
```
joaander:
```suggestion
#[derive(Clone, Copy, Debug, PartialEq)]
```
... and similarly for the other types.

Users may want to compare their shapes for equality: https://rust-lang.github.io/api-guidelines/interoperability.html?highlight=partialeq#types-eagerly-implement-common-traits-c-common-traits

Of the rest of the derivable traits on that list, I don't think we need others. Maybe Display, but deciding how to format output takes a lot of work. Most shapes don't have a concept of a `Default` -- and if they do, it doesn't derive from the defaults of the field types.

hoomd-geom/src/cuboid.rs
```diff
@@ -0,0 +1,207 @@
...
+pub struct Cuboid<const N: usize> {
+    /// The lengths of each edge of the cuboid.
+    pub edge_lengths: Cartesian<N>, // TODO: use PositiveReal
```
joaander: As with the hyperellipsoid, I don't think of the edge lengths as being represented as value in a vector space. This should be `[PositiveReal; N]`.

hoomd-geom/src/common.rs
```diff
@@ -0,0 +1,169 @@
...
+    IntersectsAt<S, Cartesian<3>, R> for HyperEllipsoid<3>
+where
+    RotationMatrix<3>: From<R>,
+{
+    #[inline]
+    fn intersects_at(&self, other: &S, v_ij: &Cartesian<3>, o_ij: &R) -> bool {
+        collide3d(self, other, v_ij, o_ij)
+    } // TODO: implement fast ellipsoid overlap
+}
+
```
joaander:
```
TODO: Implement 2D ellipse overlap.
```

hoomd-vector/src/cartesian.rs
```diff
@@ -125,6 +124,15 @@ impl<const N: usize> From<[f64; N]> for Cartesian<N> {
     }
 }
 
+impl<const N: usize> IntoIterator for Cartesian<N> {
+    type Item = f64;
+    type IntoIter = <[f64; N] as IntoIterator>::IntoIter;
+    #[inline]
+    fn into_iter(self) -> Self::IntoIter {
```
joaander, line 146: Add documentation.

hoomd-vector/src/cartesian.rs
```diff
@@ -253,6 +261,44 @@ impl<const N: usize> Add for Cartesian<N> {
...
+impl<const N: usize> Sub<f64> for Cartesian<N> {
+    type Output = Self;
+
+    #[inline]
+    fn sub(self, rhs: f64) -> Self {
+        Self {
+            coordinates: self.coordinates.map(|x| x - rhs),
+        }
+    }
+}
```
joaander: Are these used much in the new code? I don't think they really belong here. They aren't operations that mathematics defines for vector spaces, though they are convenient operations that programmers often use on arrays.

If they are used much in the code, we could implement them in `hoomd-utility` and make them more explicit with method names to remove confusion in the code.
* `elementwise_mul`
* `broadcast_add`
* `broadcast_sub`

hoomd-vector/src/cartesian.rs
joaander: Marked resolved, but not actually moved to the `Vector` trait...

hoomd-vector/src/cartesian.rs
```diff
@@ -424,6 +461,24 @@ where
     }
 }
 
+impl Cartesian<2> {
+    /// Compute a 2-vector perpendicular to self.
```
joaander: Expand documentation to mention specifically which vector it constructs.

hoomd-vector/src/cartesian.rs
```diff
@@ -890,13 +1004,40 @@ mod tests {
...
+            )
+        ]
+    )]
+    fn test_rotationmatrix_invert_3d(versor: Versor) {
+        let transposed = RotationMatrix::from(versor).inverted();
+        let inverted = RotationMatrix::from(versor.inverted());
+        for (a, b) in transposed.rows().iter().zip(inverted.rows().iter()) {
+            assert_relative_eq!(a, b);
+        }
     }
```
joaander, line 1075: Need to test:
* `Cartesian<2>::perp`
* `RotationMatrix::default`
* `RotationMatrix::Display`

hoomd-geom/src/common.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+}
+
+impl<const N: usize> SupportFn<Cartesian<N>> for HyperEllipsoid<N> {
+    #[inline]
+    fn support(&self, n: &Cartesian<N>) -> Cartesian<N> {
+        let mut denominator = self.into_iter().zip(*n).map(|(r, n)| r * n);
+        let denominator: f64 = Cartesian::<N>::from(std::array::from_fn(|_| {
+            denominator.next().unwrap_or_default()
+        }))
```
joaander: The goal here is to write code that the compiler can unroll and optimize. I don't think we need to be afraid of indexing:

```
let denominator_coordinates = array::from_fn(|i|, self.axes[i] * n[i]);
let denominator = Cartesian::from(denominator_coordinates).norm();
```
That is much more readable to my eyes. I checked Godbolt (where both axes and n are array types) and it generates tightly optimized assembly.

hoomd-geom/src/common.rs
```diff
@@ -0,0 +1,169 @@
...
+        .norm();
+        let mut iter = n
+            .into_iter()
+            .zip(self.axes)
+            .map(|(r, n)| r.powi(2) * n / denominator);
+        std::array::from_fn(|_| iter.next().unwrap_or_default()).into()
```
joaander: Should be able to rewrite this cleaner as well.

joaander (commented):
Here is a first pass at a review.

I will work on some of the obvious items, starting with merging #28 and #26 into trunk, then trunk into this branch.

We can discuss some of the higher level organization topics at the meeting today.

hoomd-geom/Cargo.toml
joaander: I see now that `itertools` provides `position_max_by`.

hoomd-geom/src/common.rs
janbridley: I fixed this in eb418b, but I'm unable to use `assert_relative_eq!` in the test. Is this expected behavior, or am I failing to import something?

joaander: Working now!

hoomd-geometry/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+// TODO: move contents of poly to shapes
+// TODO: meshes - talk to Joseph & Philipp
+// TODO: "Functions with a clear receiver are methods" for ConvexPolytope
+// TODO: implement MinDistance for spheres
+// TODO: GSD_shape_spec trait
+// TODO: impl Normals for Mesh: Mesh can have lots of cached information, with From
+//       ConvexPolytope/Polytope
+// TODO: IsInside trait (Dom)
+// TODO: scale? instead of set_volume
+// TODO: surface area, mean curvature
```
joaander: Some (most) of these things are aspirational and could be moved to issues for later PRs. I'm thinking of:
* Meshes
* GSDShape
* IsInside
* scaling
* surface area, mean curvature

hoomd-vector/src/cartesian.rs
joaander: These don't seem to be used at all. I commented them out and everything still builds and passes tests.

hoomd-geometry/src/xenocollide.rs
```diff
@@ -0,0 +1,414 @@
...
+        // Applying rotation takes ~24% of total runtime in collide3d simplex3
+        let sb_n = self
+            .q_ij
+            .rotate(&self.sb.support_mapping(&self.q_ij_inv.rotate(&n)))
+            + *self.v_ij;
+        sb_n - self.sa.support_mapping(&-n) // eq. 2.5.6 in GPG7
+    }
+    /// Create a new `SupportFunctor` from a Rotation that can be converted into a `RotationMatrix`
+    #[inline]
+    fn new<R: Rotation + Copy>(
```
joaander: I don't think that this `Copy` should be here. With the new `RotationMatrix::inverted` method, we should be able to invert the rotation matrix instead of doing another Versor/Angle->RotationMatrix conversion.

However, this change possibly impacts performance. I leave this note to remind us to check later. If we can remove this `Copy`, then we can remove it in every `IntersectsAt` implementation as well.

hoomd-geometry/src/lib.rs
janbridley: I cleaned up those TODOs, and will move them to an issue

hoomd-geom/src/cuboid.rs
janbridley: I am considering the cuboid specifically as extents along cartesian axes, so I think this makes a bit more sense?  I did change the ellipsoid case, and I'm open to changing this one

janbridley:
The docs are in a much better place, with usage examples and much better wording - I'd love another set of eyes on this!

hoomd-vector/src/cartesian.rs
```diff
@@ -356,6 +363,27 @@ impl<const N: usize> SubAssign for Cartesian<N> {
...
+    use hoomd_vector::{Cartesian, Vector};
+    let a = Cartesian::from([1.0, 2.0]);
+    let b = Cartesian::from([4.0, 0.0]);
+    let c = a.project(&b);
+    assert_eq!(c, [1.0, 0.0].into());
+    ```
+    */
+    #[inline]
+    #[must_use]
+    pub fn project(&self, b: &Self) -> Self {
```
joaander: Projection is general. This method can be moved to a default implementation in the `InnerProduct` trait.

hoomd-geometry/src/convex.rs
```diff
@@ -0,0 +1,72 @@
...
+    shape::{Circle, Sphere},
+    xenocollide::{collide2d, collide3d},
+};
+use hoomd_vector::{Cartesian, Rotate, Rotation, RotationMatrix, Vector};
+
+/// A newtype that checks for intersections using xenocollide, provided [`Convex<S>`]  implements [`SupportMapping`][`crate::SupportMapping`]
+#[derive(Clone, Debug, PartialEq)]
+pub struct Convex<S>(pub S);
+
+impl<V: Vector, S> SupportMapping<V> for Convex<S>
```
joaander: The `V: Vector` bound is not necessary.

hoomd-geometry/src/convex.rs
```diff
@@ -0,0 +1,72 @@
...
+impl<A, B, R> IntersectsAt<Convex<A>, Cartesian<2>, R> for Convex<B>
+where
+    A: SupportMapping<Cartesian<2>> + BoundingSphereRadius,
+    B: SupportMapping<Cartesian<2>> + BoundingSphereRadius,
+    R: Rotate<Cartesian<2>> + Rotation + PartialEq + Copy,
+    RotationMatrix<2>: From<R>,
+{
+    #[inline]
+    fn intersects_at(&self, other: &Convex<A>, v_ij: &Cartesian<2>, o_ij: &R) -> bool {
+        if !(Circle {
```
joaander: Perhaps a `Circle::with_radius` method will allow this code be formatted in a more readable way? It is fairly difficult to read as-is.

hoomd-geometry/src/convex.rs
```diff
@@ -0,0 +1,72 @@
...
+impl<A, B, R> IntersectsAt<Convex<A>, Cartesian<3>, R> for Convex<B>
+where
+    A: SupportMapping<Cartesian<3>> + BoundingSphereRadius,
+    B: SupportMapping<Cartesian<3>> + BoundingSphereRadius,
+    R: Rotate<Cartesian<3>> + Rotation + PartialEq + Copy,
+    RotationMatrix<3>: From<R>,
+{
+    #[inline]
+    fn intersects_at(&self, other: &Convex<A>, v_ij: &Cartesian<3>, o_ij: &R) -> bool {
+        if !(Sphere {
```
joaander: Try `Sphere::with_radius` as mentioned above for `Circle`.

hoomd-geometry/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+let s1 = Sphere {radius: 1.0};
+
+let q_id = Versor::default();
+
+// Determine the intersection between two spheres, using a fast overlap check
+assert_eq!(s0.intersects_at(&s1, &[1.9, 0.0, 0.0].into(), &q_id), true);
+assert_eq!(s0.intersects_at(&s1, &[2.1, 0.0, 0.0].into(), &q_id), false);
+
+// For more complex bodies, the `Convex` wrapper uses Xenocollide to detect overlaps
+assert_eq!(Convex(s0).intersects_at(&Convex(s1), &[1.9, 0.0, 0.0].into(), &q_id), true);
```
joaander: This example is getting ahead of itself. `Convex` is introduced in the next example.

hoomd-geometry/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+assert_eq!(Convex(s0).intersects_at(&Convex(s1), &[2.1, 0.0, 0.0].into(), &q_id), false);
+```
+By implementing the [`SupportMapping`] trait for a struct, the [`Convex`] newtype can be
+used for general, robust intersection queries between pairs of geometries. The
+[`xenocollide`] algorithm, provided for 2d and 3d shapes, is used in a blanket
+implementation on [`Convex`].
+
+```rust
+# use hoomd_geometry::{ IntersectsAt, shape::{Cuboid, Sphere} , Convex };
```
joaander: We should show the imports to that users can copy and paste the code.

hoomd-geometry/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+pub trait MinDistance<const N: usize, V, R, S> {
+    /// Minimum distance between two shapes in `N` dimensions
+    fn min_distance(&self, other: &S, v_ij: &V, o_ij: R) -> f64;
+}
+
+/**
+The support mapping of a geometry.
+
+The function associated with this trait should take a direction vector and return the
+point on a (convex) shape lying furthest in that direction.
```
joaander: I added more examples, here and elsewhere.

hoomd-geometry/src/shape/capsule.rs
```diff
@@ -0,0 +1,118 @@
...
+        assert_relative_eq!(
+            cap.volume(),
+            Hypersphere::<3> { radius }.volume()
+                + Cylinder {
+                    radius,
+                    height: cap.height
+                }
+                .volume()
+        );
+    }
```
joaander, line 269: Add tests for bounding sphere radius and support mapping.

hoomd-geometry/src/shape/capsule.rs
```diff
@@ -0,0 +1,118 @@
...
+
+use crate::{BoundingSphereRadius, SupportMapping, Volume};
+
+use hoomd_vector::{Cartesian, InnerProduct};
+
+use super::sphere::sphere_volume_prefactor;
+
+/** All points less than or equal to a distance `r` along a line of length `h`.
+This line is oriented along the `[0 0 ... 1]` direction, and has extents `+h/2`, `-h/2`
+along that axis.
```
joaander: Something is wrong with the capsule intersection tests. `intersects_at` seems to always return `false`, even when the two capsules are at the same position!

This and other tests fail.
```
let capsule = Convex(Capsule::<2> { radius: 1.0, height: 8.0 });

assert_eq!(capsule.intersects_at(&capsule, &[0.0, 0.0].into(), &Angle::identity()), true);
```

hoomd-geometry/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+pub use convex::Convex;
+
+use thiserror::Error;
+/// Enumerate possible sources of error in fallible utility methods.
+#[non_exhaustive]
+#[derive(Error, PartialEq, Debug)]
+pub enum Error {
+    /// A set of vertices is not convex.
+    #[error("Vertices do not define a convex body.")]
+    NotConvex(),
```
joaander: `ConvexPolytope` states that it is "A faceted solid defined by the convex hull of its vertices.". Therefore, it should not be an error to provide interior points.

As this error is unused, I think we should remove it. We can always add it back in when we add a `ConvexMesh` or other type that does not allow interior points.

hoomd-geometry/src/shape/convex_polytope.rs
```diff
@@ -0,0 +1,191 @@
...
+    )?;
+    # Ok(())
+    # }
+    ```
+    # Errors
+
+    * `[Error::NotConvex]` when the set of input vertices is not convex.
+    */
+    #[inline]
+    fn try_from(vertices: Vec<Cartesian<N>>) -> Result<ConvexPolytope<N>, Error> {
```
joaander: I think this would be clearer if it were named `with_vertices` instead of `try_from`.

hoomd-geometry/src/shape/convex_polytope.rs
```diff
@@ -0,0 +1,191 @@
...
+        // TODO: compute convex hull and assert convex!
+        let bounding_radius = vertices
+            .iter()
+            .map(Cartesian::norm_squared)
+            .fold(f64::NAN, f64::max)
```
joaander: Let's avoid the opportunity to easily propagate NAN through the entire simulation.

Either we make a 0-vertex polytope an error or we implicitly add the origin. I think the error would be better.

hoomd-geometry/src/shape/convex_polytope.rs
```diff
@@ -0,0 +1,191 @@
...
+# }
+```
+*/
+pub type ConvexPolyhedron = ConvexPolytope<3>;
+
+impl<const N: usize> ConvexPolytope<N> {
+    /// The vertices of the shape.
+    #[inline]
+    #[must_use]
+    pub fn vertices(&self) -> Vec<Cartesian<N>> {
```
joaander: This should return a slice.

hoomd-geometry/src/shape/convex_polytope.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    # Example
+    ```
+    use hoomd_geometry::shape::ConvexPolytope;
+
+    let equilateral_triangle = ConvexPolytope::from(3);
+    ```
+    */
+    #[inline]
+    fn from(n: usize) -> ConvexPolytope<2> {
```
joaander: This should be `regular` to clearly describe the intent of the method.

hoomd-geometry/src/shape/convex_polytope.rs
```diff
@@ -0,0 +1,191 @@
...
+        ConvexPolytope {
+            vertices: (0..n)
+                .map(|x| {
+                    let theta = std::f64::consts::PI * (x as f64) / (n as f64);
+                    Cartesian::from([f64::cos(theta), f64::cos(theta)])
```
joaander:
```suggestion
                    Cartesian::from([f64::cos(theta), f64::sin(theta)])
```
How did this ever pass tests?

hoomd-geometry/src/shape/convex_polytope.rs
```diff
@@ -0,0 +1,191 @@
...
+        assert_eq!(simplex3.bounding_radius, f64::sqrt(3.0));
+        assert_eq!(equilateral_triangle.bounding_radius, f64::sqrt(1.0));
+    }
+
+    #[rstest]
+    fn test_bounding_radius_regular_polygons(#[values(1, 3, 8, 64)] n: usize) {
+        assert_eq!(ConvexPolygon::from(n).bounding_radius, 1.0);
+        assert_eq!(ConvexPolytope::from(n).bounding_radius, 1.0);
+    }
+}
```
joaander, line 765: Test SupportMapping.

hoomd-geometry/src/shape/cuboid.rs
```diff
@@ -0,0 +1,235 @@
+// Copyright (c) 2024-2025 The Regents of the University of Michigan.
+// Part of hoomd-rs, released under the BSD 3-Clause License.
+
+/*! Implement [`Cuboid`] */
+
+use crate::{BoundingSphereRadius, SupportMapping, Volume};
+use hoomd_vector::Cartesian;
+use itertools::multizip;
+
+/** The Minkowski sum of `N` mutually perpendicular vectors `[0... edge_lengths[i] ...0]`.
```
joaander: I get that this defines a cuboid without saying cuboid, but this seems a little too abstract. However, I don't have a better way to describe an N-dimensional rectangular cuboid.

Also, with the vectors defined as they are, the origin of the cuboid would be in the lower left, but the implementation places the origin at the centroid.

hoomd-geometry/src/shape/cuboid.rs
```diff
@@ -0,0 +1,235 @@
...
+
+```
+*/
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Cuboid<const N: usize> {
+    /// The lengths of each edge of the cuboid.
+    pub edge_lengths: Cartesian<N>,
+}
+
+/// The Minkowski sum of the vectors `[0, edge_lengths[0]]` and `[edge_lengths[1], 0]`.
```
joaander: A rectangle is such a universal concept that I think it is OK to use its name to define itself.

hoomd-geometry/src/shape/cuboid.rs
```diff
@@ -0,0 +1,235 @@
...
+    #[must_use]
+    pub fn a(&self) -> f64 {
+        self.edge_lengths[0]
+    }
+    /// Length of the `Cuboid` edge along the y axis
+    #[inline]
+    #[must_use]
+    pub fn b(&self) -> f64 {
+        self.edge_lengths[1]
+    }
```
joaander: Should there be corresponding implementations for `Cuboid<2>`?

hoomd-geometry/src/shape/cuboid.rs
```diff
@@ -0,0 +1,235 @@
...
+            .into_iter()
+            .reduce(|acc, x| acc * x)
+            .unwrap_or(0.0)
+    }
+}
+
+impl<const N: usize> BoundingSphereRadius for Cuboid<N> {
+    #[inline]
+    fn bounding_sphere_radius(&self) -> f64 {
+        f64::sqrt(3.0) / 2.0 * self.edge_lengths.into_iter().fold(f64::NAN, f64::max)
```
joaander: Double check your implementation. I do not get expected bounding radii in the documentation tests that I wrote.

hoomd-geometry/src/shape/cuboid.rs
```diff
@@ -0,0 +1,235 @@
...
+
+// TODO: requires test
+impl<const N: usize> SupportMapping<Cartesian<N>> for Cuboid<N> {
+    #[inline]
+    fn support_mapping(&self, n: &Cartesian<N>) -> Cartesian<N> {
+        let mut iter = n
+            .into_iter()
+            .zip(self.edge_lengths)
+            .map(|(n_i, l_i)| l_i / 2.0 * n_i.signum());
+        std::array::from_fn(|_| iter.next().unwrap_or_default()).into()
```
joaander:
```suggestion
        std::array::from_fn(|_| iter.next().expect("N should match on the input and output").into()
```

hoomd-geometry/src/shape/cuboid.rs
```diff
@@ -0,0 +1,235 @@
...
+
+// The extent along the z-axis is different for the two example cuboids
+assert_ne!(min_extents[2], rectangular_prism.minimal_extents()[2]);
+
+```
+*/
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Cuboid<const N: usize> {
+    /// The lengths of each edge of the cuboid.
+    pub edge_lengths: Cartesian<N>,
```
joaander: I still think that it is not valid to represent this with a Cartesian vector. Given that the origin is at the center, we can't even say that this vector points to the corner of the box or something like that. I converted this to `[f64; N]`.

We should consider making it `[PositiveReal; N]`. That would make initialization more annoying, but would prevent the user from accidentally triggering degenerate cases when providing negative values. Let's leave it for now and consider the `PositiveReal` usage throughout this crate (sphere radii, capsule height, etc..) for a possible future PR if we decide that it is worth the hassle.

hoomd-geometry/src/shape/cuboid.rs
```diff
@@ -0,0 +1,235 @@
...
+        );
+    }
+
+    #[rstest(
+        l => [1e-6, 1.0, 3.456, 99_999_999.9],
+    )]
+    fn test_box_abc(l: f64) {
+        let c = Cuboid::from([l; 3]);
+        assert_eq!([c.a(), c.b(), c.c()], [l; 3]);
+    }
```
joaander, line 482: Need to test both SupportMapping and BoundingSphereRadius.

hoomd-geometry/src/shape/hyperellipsoid.rs
```diff
@@ -0,0 +1,414 @@
...
+use std::ops::{Add, Mul};
+
+/// TODO: temp, remove once we have a linalg crate
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct SquareMatrix<const N: usize> {
+    /// The elements of the matrix
+    pub(crate) rows: [[f64; N]; N],
+    // diagonal: bool,
+    // symmetry: ???
+}
```
joaander:
```suggestion
#[derive(Clone, Copy, Debug, PartialEq)]
struct SquareMatrix<const N: usize> {
    /// The elements of the matrix
    rows: [[f64; N]; N],
    // diagonal: bool,
    // symmetry: ???
}
```
Why are these public? They don't seem to be used outside this module.

hoomd-geometry/src/shape/convex_polytope.rs
```diff
@@ -0,0 +1,191 @@
...
+// Part of hoomd-rs, released under the BSD 3-Clause License.
+
+/*! N-Dimensional generalization of a convex polyhedron.*/
+use crate::{BoundingSphereRadius, Error, SupportMapping};
+use hoomd_vector::{Cartesian, InnerProduct};
+
+/**A faceted solid defined by the convex hull of its vertices.*/
+pub struct ConvexPolytope<const N: usize> {
+    /// The vertices of the shape.
+    vertices: Vec<Cartesian<N>>,
```
joaander, line 59: We should provide a way for callers to mutate the vertices. I'm not sure if it is worth mutating them in place, or just provide a `set_vertices` call that rewrites all the verts.

hoomd-geometry/src/shape/sphere.rs
```diff
@@ -0,0 +1,195 @@
...
+pub(crate) fn factorial(n: usize, ntuple: usize) -> usize {
+    assert!(ntuple > 0);
+    if n == 0 {
+        1
+    } else {
+        (1..=n)
+            .rev()
+            .step_by(ntuple)
+            .reduce(|acc, x| acc * x)
+            .unwrap_or_default() // inaccessible: 1..=(n!=0) is never empty
```
joaander: This is exactly what `expect` is for...

hoomd-geometry/src/shape/sphere.rs
```diff
@@ -0,0 +1,195 @@
+// Copyright (c) 2024-2025 The Regents of the University of Michigan.
+// Part of hoomd-rs, released under the BSD 3-Clause License.
+
+/*! Implement [`Hypersphere`] */
+use crate::{BoundingSphereRadius, IntersectsAt, SupportMapping, Volume};
+use hoomd_vector::{InnerProduct, Rotate, Vector};
+use std::f64::consts::PI;
+
+/// The (single, double, ...)-factorial function
+pub(crate) fn factorial(n: usize, ntuple: usize) -> usize {
```
joaander: I'm removing some `pub(crate)` that are unused.

hoomd-geometry/src/shape/sphere.rs
```diff
@@ -0,0 +1,195 @@
...
+    #[case(PhantomData::<Hypersphere<4>>)]
+    #[case(PhantomData::<Hypersphere<5>>)]
+    fn test_support_fn<const N: usize>(
+        #[case] _n: PhantomData<Hypersphere<N>>,
+        #[values(0.1, 1.0, 33.3)] radius: f64,
+    ) {
+        let s = Hypersphere::<N> { radius };
+        let v = Cartesian::<N>::from([radius.powi(2) / 1.8; N]);
+        assert_eq!(v / v.norm() * radius, s.support_mapping(&v));
+    }
```
joaander, line 300: Need to test `intersects_at`.

hoomd-geometry/src/xenocollide.rs
```diff
@@ -0,0 +1,425 @@
...
+use crate::SupportMapping;
+use hoomd_vector::{Cartesian, Cross, InnerProduct, Rotate, Rotation, RotationMatrix};
+
+/// Maximum allowed iterations for Xenocollide in 2D
+const XENOCOLLIDE_2D_MAX_ITER: usize = 1024;
+/// Maximum allowed iterations for Xenocollide in 3D
+const XENOCOLLIDE_3D_MAX_ITER: usize = 1024;
+
+/// Stateful function for support function calculations on Minkowski differences.
+struct SupportFunctor<
```
joaander: I don't think we need to call things "Functor"s in Rust. Let's rename it to `MinkowskiDifference`.

hoomd-geometry/src/xenocollide.rs
```diff
@@ -0,0 +1,425 @@
...
+        // If the origin is outside the support plane, return false (no overlap)
+        if v3.dot(&v_perp_v2v1) < 0.0 {
+            return false;
+        }
+
+        // Tolerance check. NOTE: may not always be necessary?
+        let d = (v3 - v1) - (v3 - v1).project(&(v2 - v1)) * tol_multiplier;
+        if d.norm_squared() < tol.powi(2) * v3.norm_squared() {
+            return true;
+        }
```
joaander: Certainly not always necessary, but sometimes yes. The HOOMD-blue code documents one case where using no tolerance leads to infinite loops.

You missed some parens, not sure how this is even working.
```suggestion
        // Tolerance check. NOTE: may not always be necessary?
        let d = ((v3 - v1) - (v3 - v1).project(&(v2 - v1))) * tol_multiplier;
        if d.norm_squared() < tol.powi(2) * v3.norm_squared() {
            return true;
        }
```
Actually, I don't think the whole `tol_multipler` thing is necessary. I will remove it.

hoomd-geometry/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// The N-hypervolume of a geometry.
+    #[must_use]
+    fn volume(&self) -> f64;
+}
+
+/**
+Definitions of the minimum distance between two shapes. Will be zero if points are on
+a boundary (within floating-point precision) and negative if the shapes are overlapping.
+*/
+pub trait MinDistance<const N: usize, V, R, S> {
```
joaander: Should we just remove MinDistance for now until we decide how we want to implement it?

joaander (changes requested):
Thanks! This is so close to being done.

I made a lot of comments and implemented the fixes for most of them directly. I will resolve these. 

There are a few unresolved comments, including a few bugs that need to be investigated (in Capsule and Cuboid) and some possible unsafe behavior. There is also a lot more testing that needs to be done. 

In the interests of getting this crate merged in and ready to use, I'm would like to open TODO issues to complete those tasks in future PRs.

There is one thing we should discuss and probably decide on before merging. Should we use `PositiveReal` for `Hypersphere`, `Hyperellipsoid`, `Cuboid`, `Cylinder`, `Capsule`, and `Sphero` parameters? All of these assume that their parameters are positive (though some may work with 0). Providing negative values will lead to undefined behavior. Using `PositiveReal` (or `NonNegativeReal`) would make these types slightly more difficult to construct, but we would prevent the user from accidentally triggering undefined behavior and potentially save them many hours of debugging. If you all agree, I'd like to make this change before merging so that we don't have to go and add `.into()?` in a thousand places that use this crate if we decide to do it later.

hoomd-geometry/src/shape/convex_polytope.rs
joaander: We should translate the tests from the HOOMD-blue source code.

janbridley:
> Thanks! This is so close to being done.
> 
> 
> 
> I made a lot of comments and implemented the fixes for most of them directly. I will resolve these. 
> 
> 
> 
> There are a few unresolved comments, including a few bugs that need to be investigated (in Capsule and Cuboid) and some possible unsafe behavior. There is also a lot more testing that needs to be done. 
> 
> 
> 
> In the interests of getting this crate merged in and ready to use, I'm would like to open TODO issues to complete those tasks in future PRs.
> 
> 
> 
> There is one thing we should discuss and probably decide on before merging. Should we use `PositiveReal` for `Hypersphere`, `Hyperellipsoid`, `Cuboid`, `Cylinder`, `Capsule`, and `Sphero` parameters? All of these assume that their parameters are positive (though some may work with 0). Providing negative values will lead to undefined behavior. Using `PositiveReal` (or `NonNegativeReal`) would make these types slightly more difficult to construct, but we would prevent the user from accidentally triggering undefined behavior and potentially save them many hours of debugging. If you all agree, I'd like to make this change before merging so that we don't have to go and add `.into()?` in a thousand places that use this crate if we decide to do it later.

I do think it would be a good idea to use PositiveReal where possible - the reduction in bugs is well worth the additional syntax on construction. I'd also recommend PositiveReal instead of NonNegative real even if the latter is technically valid, as it provides a stronger guarantee that our methods are correct. If you don't beat me to it, I'll work on that on Monday!

joaander:
> If you don't beat me to it, I'll work on that on Monday!

Great! I've been focusing on the other unit test for now. I'll leave the validated fields up to you then. I mentioned `NonNegative` because I noticed that the capsule uses a 0 height in some unit tests. Some of the Cuboid tests set a 0 edge length. 

You have a point about requiring `PositiveReal` for all of these. I don't see any problems requiring that these shapes have a non-zero volume. Degenerate shapes (i.e. 0-radius capsules) are likely better represented with a purpose-built shape (i.e. line segment) where we are sure that the shape is implemented correctly. I guess the same goes for a degenerate capsule with height=0. Say we someday implement the fast intersection check - that may very well be unstable when the height is 0 unless we take care to handle that special case.

You'll need to adjust or replace the unit tests appropriately that currently set 0's.

hoomd-geometry/src/shape/convex_polytope.rs
joaander: I tried to implement this, but there are too many open questions about how it might be used. Let's wait until we have a concrete use-case and can then design an appropriate API.

joaander (approved):
I implemented the `PositiveReal` changes. This branch is now ready to merge!

janbridley:
> I implemented the `PositiveReal` changes. This branch is now ready to merge!

Looks great! I'm excited to have these in.

PR #11, opened 2025-04-17
Implement Microstate and MC

joaander:
Prepare the Microstate and MC frameworks.

Implement the frameworks for bodies with open boundary conditions and external potentials. After we merge this PR, we can begin implementing other pieces of the framework a little at a time in more focused PRs. I will create issues to track the work that needs to be done.

Todo before merging:
- [x] Implement Ball distribution.
- [x] Free fall example
- [x] Unit test Microstate
- [x] Document Microstate
- [x] Unit test MC
- [x] Document MC


joaander:
Overview documentation for `hoomd-interaction` and `hoomd-mc` will wait until there is enough content to provide an overview. Specifically, I'm thinking that square well and/or LennardJones pairwise interactions is the minimum level of complexity for a reasonable introduction.

hoomd-interaction/src/lib.rs
```diff
@@ -8,7 +8,193 @@
...
+    a: f64,
+    b: f64,
+}
+
+impl<S> SiteEnergy<S> for Custom
+where
+    S: Position<Cartesian<2>>
+{
+    fn site_energy(&self, site_properties: &S) -> f64 {
+        self.a * (site_properties.position()[0] / self.b).cos()
```
janbridley, line 66:
```suggestion
        self.a * f64::cos(site_properties.position()[0] / self.b)
```

joaander: This works, but is it idiomatic? All the examples in the std library docs I find use the `value.cos()` syntax.

PR #21, opened 2025-05-13
Boundary conditions

joaander:
Implement the `Boundary` trait and the `Square` boundary.

* Methods that add or update bodies/sites are now fallible.
* `Sweep` rejects trial moves that leave the boundary.
* `Microstate` is now generic on the vector space `V` via `PhantomData`.

Boundaries are not "walls". They apply only to the positions of the bodies and sites.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -34,7 +34,6 @@ jobs:
           # - macos-14
         rust:
           # Test multiple rust versions on multiple platforms in debug mode.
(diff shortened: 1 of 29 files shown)


joaander:
Question: Is `Error::CannotWrapProperties` sufficient for `add_body`, `extend_body`, and `update_body_properties`? Or would you want a different error variant to determine whether the body properties or some site property caused the error?

PR #22, opened 2025-05-14
Add harmonic potential implementation

shihkual:
Add harmonic potential as an external field and an isotropic pair potential (maybe bonding potential?)


joaander:
I would implement this only as a pairwise potential. I don't see any users needing a potential that bonds every site to a single global point via harmonic bonds.

The pairwise code can be repurposed for cases like Frenkel-Ladd where each site has a reference position that it is bonded to. Users would implement this in custom code that implements `SiteEnergy` and temporarily forms a `pairwise::Harmonic` to compute the energy between the site and its reference.

`SitePairEnergy` (coming soon) will be generalized to work with all potentials, including `Harmonic`.

shihkual:
Yes, I totally agree. Implementing it as either an external field or a pure pairwise potential among all the particles initially seems strange to me. I started by prototyping it as an external field to become familiar with the hoomd-rs code.

I can temporarily prototype a `pairwise::Harmonic` that calculates the harmonic potential based on distance and could adapt it to the new `SitePairEnergy` later on.

joaander:
Thanks!

joaander:
Also, if you have time the `HarmonicRepulsion` is related to `Harmonic`, but set to 0 for r > r_0 (Similar to how WeeksChandlerAnderson is LJ cutoff at 2**(1/6) sigma). `HarmonicRepulsion` (called `dpd_conservative` in HOOMD-blue) is useful as a soft potential that penalizes overlaps.

shihkual:
Will do!

shihkual:
I have a question related to how Rust works. I implemented the harmonic potential defined by spring constant `k` and equilibrium spring length `r0`. I want the `k` to be input by the user explicitly, and `r0` defaults to 0.0. The current implementation will still need to provide `r0` explicitly as:
```
let harmonic = Harmonic::new(1.0, None)
```
or 
```
let harmonic = Harmonic::new(1.0, Some(1.0))
```
Is there a better way to implement it, or is this a formal way to set a default input?

joaander:
> I have a question related to how Rust works. I implemented the harmonic potential defined by spring constant `k` and equilibrium spring length `r0`. I want the `k` to be input by the user explicitly, and `r0` defaults to 0.0. The current implementation will still need to provide `r0` explicitly as:
> 
> ```
> let harmonic = Harmonic::new(1.0, None)
> ```
> 
> or
> 
> ```
> let harmonic = Harmonic::new(1.0, Some(1.0))
> ```
> 
> Is there a better way to implement it, or is this a formal way to set a default input?

Sorry I missed this. There are a two common ways in Rust way to provide default values for some fields and defaults for others. One looks like this:

```
let harmonic = Harmonic { k=1.0, ..Default::default() }
```
To make this work, you need to `impl Default for Harmonic`. This doesn't exactly meet all your requirements because `Harmonic::default()` must also set a default value for `k`.

The alternate method is called the "Builder" pattern and it allows you to set some values as overridable defaults but require others. The main builder pattern is described here: https://rust-lang.github.io/api-guidelines/type-safety.html?highlight=builder#builders-enable-construction-of-complex-values-c-builder. 

A full builder pattern harmonic would look like:
```
let harmonic = HarmonicBuilder::with_k(1.0).build();
```
or
```
let harmonic = HarmonicBuilder::with_k(1.0)
    .r0(2.0)
    .build();
```

The ratatui library implements what it calls the "Builder-Lite" pattern which might be better for simple structs like `Harmonic`: https://ratatui.rs/concepts/builder-lite-pattern/ . I haven't worked with this pattern enough to know how it would look here.

In your case, I'm not sure the `Option` makes sense for `r0`. What does the absence of a value for `r0` mean? Take the equation literally: `1/2 k * (r - r_0)**2` and it is clear that there must always be an r0. With simple structs that have all public members, I think we should even skip the `new` methods entirely and ask the user to construct them directly:

```
let harmonic = Harmonic { k=1.0, r0: 0.0 }
```
This is simple, explicit, and offers no room for confusion.

It does take a little used to coming from C++ and Python where default arguments are commonly used. One think I like about the struct initialization is that it looks somewhat like Python keyword arguments which help to make it clear what numbers mean what when there is a large number of fields to set.

shihkual:
I see. Yeah, this makes more sense to me.
```
let harmonic = Harmonic { k=1.0, r0: 0.0 }
```
Looks more concise and intuitive. Will make changes.

Both options make sense to me. We can either assign a default value to `r0` or leave it unset. I was considering the use case of the Frenkel-Ladd method, where you bind the particles to their lattice sites, so `r0` is always 0. However, for the bonding potential, `r0` should be some finite value, so I was pondering whether to assign it a default value or not.

I noticed that the way we currently implement the pair potential is by always using the `new` methods, so I planned to follow this pattern for consistency previously. Are we planning to re-formulate all of them?

joaander:
> Are we planning to re-formulate all of them?

Yes, I did so in #26. We can always revisit this as there are pros and cons on both sides. My current thinking is that the `new` methods are redundant when the struct has all public fields. The alternative would be to make the fields private, provide `new`, getter, and setter methods. That is a lot of boilerplate to provide for each potential.

shihkual:
@joaander Ready for review. Need some help to solve the pre-commit issue.

joaander (approved):
Thanks!

I cherry-picked the commit that fixes the equations. I also pushed a few revisions to the documentation directly to the branch. The biggest change I made was to rewrite the `HarmonicRepulsion` terms in a mathematically equivalent form that clearly shows the relation to the harmonic spring potential/force.

PR #26, opened 2025-05-21
Cutoff pair

### LINKED ISSUE #14: Implement CutoffPair in hoomd-interaction
The `SitePairEnergy` trait will go along with `CutoffPair` to evaluate the energy of a pair of sites. 

* `CutoffPair` will implement the for loops needed to sum the energies, calling to an inner type that implements `SitePairEnergy` to evaluate the energy.
* To generalize over all isotropic pair energies, the `Isotropic` newtype will wrap a type that implements `IsotropicEnergy`.
* To generalize over all anisotropic eneriges, the `Anisotropic` type newtype will wrap a type that implements `AnisotropicEnergy`.

To allow `CutoffPair` to work for types that implement both energy and force as well as types that only implement one of the two, we will need to generalize the `energy` and `force` methods with the appropriate trait bounds (or at the impl level: what is the difference?).

We also need to implement `TotalEnergy` and `DeltaEnergyOne` for `CutoffPair`.

joaander:
Implement `CutoffPair` and clean up many TODO items in hoomd-interaction.

Resolves #14/


hoomd-interaction/src/pairwise/xplor.rs
```diff
@@ -103,8 +86,8 @@ impl<F: IsotropicForce + IsotropicEnergy> IsotropicForce for Xplor<F> {
         } else if r > self.r_cut {
             0.0
         } else {
-            // Chain rule of s(r)*U(r)
-            self.s(r) * self.f.force(r) + self.ds_dr(r) * self.f.energy(r)
+            // Chain rule of -d/dr(s(r)*U(r))
+            self.s(r) * self.f.force(r) - self.ds_dr(r) * self.f.energy(r)
```
joaander, line 91: @janbridley I implemented the missing unit test that checks the XPLOR force against a separately calculated quantity. That test found a sign error here.

joaander:
Merging now so we can sync up with #6 and make forward progress. The remaining TODO items can be handled in a later PR:
- Test `iter_sites_near`
- Test energy sum implementations for `TotalEnergy` and `DeltaEnergy`.
- Consider moving `DeltaEnergy` to hoomd-interaction.

hoomd-interaction/src/pairwise/xplor.rs
janbridley: Sounds great - thank you!

PR #28, opened 2025-05-27
Use Katex

joaander:
Replace all of the MathML with Katex rendering on the client side.

Also enable readthedocs builds.


joaander:
@josephburkhart Could you look at the javascript in `katex.html` and make any suggestions for improvement. The goal is to support GitHub flavored LaTeX delimiters (the codeblock variants). This stackoverflow explains the goals in detail: https://stackoverflow.com/questions/48438067/remove-text-from-around-dom-node

I found several alternate implementations:
* https://github.com/CAD97/katex-doc/blob/master/katex.html
* https://github.com/m-ou-se/rust-horrible-katex-hack/blob/565ffd921d3c0eef2037aef58215c56ba1a09ddc/src/lib.rs
* https://github.com/rust-lang/rust/pull/95691#issuecomment-1089080597

The third is a newer version of the second by the same author.

The alternative is to use `renderMathInElement` directly from KaTeX (https://github.com/paulkernfeld/rustdoc-katex-demo/blob/master/katex-header.html). This, however, will require that we use backslashes to escape `_`, `*`, and other special markdown characters. Keeping all math in code blocks prevents this (https://docs.rs/rustdoc-katex-demo/latest/rustdoc_katex_demo/)

PR #31, opened 2025-06-03
Relaxed conditions in hoomd-vector to allow for vector spaces with no inner product

mthran:
To accommodate Minkowski and curved vector spaces, the methods `dot`, `norm`, `norm_squared`, `to_unit`, and `to_unit_unchecked` have been moved to a new subtrait `InnerProduct` of `Vector`. With this classification, `Vector` describes vectors living in a metric space and supports normal vector operations (vector addition, vector negation, and rescaling), as well as the new method `distance` which computes the distance between two vectors using a given metric. The subtrait `InnerProduct` implements the same methods as the older version of the trait `Vector`. 

`Vector` is still used to define position vectors in cases where norms and dot products are not needed (e.g., as in `Microstate`, `Translate`, `SiteEnergy`). The `Linear` external potential and the `AnisotropicEnergy` for the `AngularMask` pairwise potential require the more restrictive `InnerProduct` subtrait.


hoomd-vector/src/lib.rs
```diff
@@ -336,6 +340,42 @@ pub trait Vector:
     + SubAssign
     + Neg<Output = Self>
 {
+    /** Compute the squared distance between two vectors belonging to a metric space.
+     */
+    fn distance_sq(&self, other: &Self) -> f64;
+
+    /** Compute the distance between two vectors belonging to a metric space.
+     */
+    fn distance(&self, other: &Self) -> f64;
```
joaander: Does the distance metric in curved space end up computing the square root of a distance squared? If so, we can save a few lines of code by providing a default implementation. If not, the default can always be overridden.

```suggestion
    #[inline]
    fn distance(&self, other: &Self) -> f64 {
        self.distance_sq(other).sqrt()
    }
```
After this change, we can remove the implementation of `distance` from `Cartesian<N>`.

hoomd-vector/src/lib.rs
```diff
@@ -336,6 +340,42 @@ pub trait Vector:
     + SubAssign
     + Neg<Output = Self>
 {
+    /** Compute the squared distance between two vectors belonging to a metric space.
+     */
+    fn distance_sq(&self, other: &Self) -> f64;
```
joaander:
```suggestion
    fn distance_squared(&self, other: &Self) -> f64;
```
I'm OK with `_sq` if others have preference for that. If we choose to keep `_sq` then we should rename `norm_squared` to `norm_sq` for consistency.

hoomd-vector/src/cartesian.rs
```diff
@@ -228,14 +228,43 @@ impl<const N: usize> TryFrom<[f64; N]> for Unit<Cartesian<N>> {
...
+    d^2(\vec{x},\vec{y}) = \sum_{i=1}^{N} (x_i - y_i)^2
+    ```
+
+    # Example
+    ```
+    use hoomd_vector::{Cartesian, Vector};
+
+    # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let x = Cartesian::<3>::try_from([0.0, 1.0, 1.0])?;
+    let y = Cartesian::<3>::try_from([1.0, 0.0, 0.0])?;
```
joaander:
```suggestion
    let x = Cartesian::from([0.0, 1.0, 1.0]);
    let y = Cartesian::from([1.0, 0.0, 0.0]);
```
Rust can look at the length of the array to determine the ::<3>. `try_from` is also not needed for arrays.

hoomd-vector/src/lib.rs
```diff
@@ -336,6 +340,42 @@ pub trait Vector:
     + SubAssign
     + Neg<Output = Self>
 {
+    /** Compute the squared distance between two vectors belonging to a metric space.
+     */
```
joaander, line 364:
```suggestion

    # Example
    ```
    use hoomd_vector::{Cartesian, Vector};

    # fn main() -> Result<(), Box<dyn std::error::Error>> {
    let x = Cartesian::from([0.0, 1.0, 1.0]);
    let y = Cartesian::from([1.0, 0.0, 0.0]);
    assert_eq!(3.0, x.distance_sq(&y));
    # Ok(())
    # }
    ```
     */
```
Let's repeat the Cartesian example here. That way users will see how to use the method even if they discover it in the top level docs.

hoomd-vector/src/lib.rs
```diff
@@ -336,6 +340,42 @@ pub trait Vector:
...
+    fn distance_sq(&self, other: &Self) -> f64;
+
+    /** Compute the distance between two vectors belonging to a metric space.
+     */
```
joaander, line 376:
```suggestion
    # Example
    ```
    use hoomd_vector::{Cartesian, Vector};

    let x = Cartesian::from([0.0, 0.0]);
    let y = Cartesian::from([3.0, 4.0]);
    assert_eq!(5.0, x.distance(&y));
    ```
     */
```

hoomd-vector/src/lib.rs
```diff
@@ -17,25 +17,28 @@ other fields may find `hoomd_vector` useful outside the context of `HOOMD`.
 
 ## Vectors
 
-The [`Vector`] trait describes any type that is a member of a normed vector
+The [`Vector`] trait describes any type that is a member of a metric vector
 space. Write code with a [`Vector`] trait bound when you can express the
-computation with vector arithmetic and dot products. Your generic code can
+computation with vector arithmetic and a distance metric. Your generic code can
 then be invoked on vector types with any dimension or representation (e.g.
 spherical coordinates).
```
joaander, line 25: This and the `some_other_function` later are begging for an example:
```suggestion
use hoomd_vector::Vector;

fn some_function<V: Vector>(a: &V, b: &V, c: &V) -> f64 {
    (*a + *b).distance(&c)
}
```
Note: need to surround the example in triple backticks. GitHub doesn't parse them correctly when I try to add them here.

joaander (commented):
Great, thanks for this! I have few suggestions for improvement.

joaander (approved):
Thanks!

PR #32, opened 2025-06-30
Basic infrastructures for chimes interatomic potential

shihkual:
Build basic utility codes and prototype the ChIMES interatomic potential, implemented natively in hoomd-rs.

Todo:


hoomd-interaction/src/pairwise.rs
```diff
@@ -31,6 +31,15 @@ pub use harmonic::Harmonic;
 mod harmonic_repulsion;
 pub use harmonic_repulsion::HarmonicRepulsion;
 
+mod chimes_cheby2b;
```
joaander: I would put these in the `hoomd-chimes` crate. That keeps things more independent so that a user needs to only pull in `hoomd-interaction` to use the basic potentials and can add `hoomd-chimes` only when needed.

In other words, `hoomd-chimes` should depend on `hoomd-interaction`, but not the other way around.

shihkual: Gotcha. Yeah good point. I will move the scripts.

hoomd-chimes/src/cheby.rs
```diff
@@ -0,0 +1,190 @@
...
+
+/** Evaluates the Chebyshev polynomials and its derivatives
+  of the first kind $`T_i(s)`$ for orders $`i`$ equals 0 to
+  $`n - 1`$, given coordinate $`s`$.
+
+*/
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Chebyshev {
+    /// Maximum order of the Chebyshev polynomials $`T_i`$ (i = 0 to n-1).
+    pub n: usize,
```
joaander: Does `n` need to be determined at runtime? If you can set N at compile time, make it a const generic. Then the evaluation functions below can be made associated functions and called via `Chebyshev::<3>::f(s)` for example. In that case, the Type seems unnecessary. You could just as easily have the evaluation functions be module level functions that take a const generic N:
```
chebyshev::t<N>(s)
```

hoomd-chimes/src/cheby.rs
```diff
@@ -0,0 +1,190 @@
...
+    ```
+
+    # Notes
+
+    If `n` is 0, an empty vector is returned.
+    */
+    #[must_use]
+    #[inline]
+    pub fn eval_cheby(&self, s: &f64) -> Vec<f64> {
+        let mut tn: Vec<f64> = vec![0.0; self.n];
```
joaander: A new memory allocation/deallocation multiple times for every pair potential evaluation will be insanely expensive. If you can set N at compile time, this can be a fixed sized array allocated on the stack.

If `n` must be determined at runtime then the best you can do is have the `evaluate` function take a mutable slice and populate that with results. `n` would be determined by the length of the slice. The caller can then allocate one vec and reuse it for multiple evaluations.

Also, the `cheby` in this name is redundant: `Chebyshev::eval_cheby`. It should be called `evaluate` or `t`.

hoomd-chimes/src/cheby.rs
```diff
@@ -0,0 +1,190 @@
...
+
+        if self.n == 0 {
```
joaander: This is ill-defined, so it should panic instead of returning a nonsense value.

hoomd-chimes/src/cheby.rs
```diff
@@ -0,0 +1,190 @@
...
+    assert_eq!(tnd[3], 0.0); // dT_3/ds = 3 * U_2 = 3 * (4(0.5)^2 - 1) = 0.0
+    ```
+
+    # Notes
+
+    If `n` is 0, an empty vector is returned.
+    */
+    #[must_use]
+    #[inline]
+    pub fn eval_dcheby_ds(&self, s: &f64) -> Vec<f64> {
```
joaander: This could similarly be named `evaluate_derivative`.

Alternately, the two methods could be `fn t()` and `fn dt_ds`.

hoomd-chimes/src/potential/chimes_cheby2b.rs
```diff
@@ -0,0 +1,191 @@
...
+Chebyshev polynomials, and $`s`$ is the transformed
+distance between particles, given by [`Transformation`].
+
+# Note:
+* See equation 2 in <https://doi.org/10.1038/s41524-024-01497-y>.
+* Must be used with the [`TersoffSmooth`] and [`ChimesPenalty`]
+to enable correct potential calculation.
+ */
+#[derive(Clone, Debug, PartialEq)]
+pub struct Chimes2b<F: Transformation, T = Chebyshev> {
```
joaander: I see. You are using `Chebyshev` as a type so that this method can be made generic for other basis sets. In that case, I would recommend that the methods be `evaluate` and `evaluate_derivative`. To allow for generic use, they will need to be defined in a base trait that is implemented by `Chebyshev`.

hoomd-chimes/src/potential/chimes_expansion.rs
```diff
@@ -0,0 +1,191 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub struct Chimes2b<F: Transformation, T = Chebyshev> {
+    /// Transformation style.
+    pub trans_style: F,
+    /// one plus two-body `ChIMES` coefficient (`[energy]`).
+    pub coeff: Vec<f64>,
+    /// Inner radial cut-off (`[length]`).
+    pub r_in: f64,
+    /// Buffer distance before triggering the damping.
+    inner_smooth_r: f64,
```
joaander, line 36: clippy strongly recommends that all fields be public or all fields be private. If you need to have some private fields, make all the others private as well and add accessor methods.

hoomd-chimes/src/potential/chimes_penalty.rs
```diff
@@ -0,0 +1,152 @@
...
+impl IsotropicEnergy for ChimesPenalty {
+    #[inline]
+    fn energy(&self, r: f64) -> f64 {
+        let r_penalty = self.r_in + self.dt - r;
+
+        if r_penalty <= 0.0 {
+            0.0
+        } else {
+            let e_penalty = self.a * r_penalty * r_penalty * r_penalty;
+            println!(
```
joaander: There should be no printlns in a production library.

joaander (commented):
Sorry for the delay in reviewing this. I have some more suggestions to improve the code. I don't know what all the use-cases are for this code or what future generalizations you are planning, so some of my suggestions may not apply.

hoomd-chimes/src/potential/chimes_penalty.rs
shihkual: @joaander Thanks! I just saw the comments. Those are good ones. I will see if I can integrate them into my code.

PR #33, opened 2025-07-02
LJ-Gauss

mthran:
Implements the Lennard-Jones-Gauss pairwise potential


hoomd-interaction/src/pairwise/lennard_jones_gauss.rs
```diff
@@ -0,0 +1,89 @@
...
+pub struct LennardJonesGauss {
+    /// Scale of Gaussian
+    pub epsilon: f64,
+    /// Width of Gaussian
+    pub sigma_squared: f64,
+    /// Gaussian center
+    pub r_0: f64,
+}
+
+impl Default for LennardJonesGauss {
```
joaander: Is `r_0 = 0` the most common way that this potential will be used?

If not, it OK to omit `impl Default` entirely.

hoomd-interaction/src/pairwise/lennard_jones_gauss.rs
```diff
@@ -0,0 +1,89 @@
...
+use hoomd_interaction::pairwise::{LennardJonesGauss};
+
+let mut lennard_jones_gauss: LennardJonesGauss = LennardJonesGauss::default();
+lennard_jones_gauss.epsilon = 1.5;
+lennard_jones_gauss.sigma_squared = 3.0;
+```
+*/
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct LennardJonesGauss {
+    /// Scale of Gaussian
```
joaander: Document units of each field.

hoomd-interaction/src/pairwise/lennard_jones_gauss.rs
```diff
@@ -0,0 +1,89 @@
...
+
+/*! Implement [`LennardJonesGauss`]
+ */
+
+use super::{IsotropicEnergy, IsotropicForce};
+
+/** Double-well potential with a steep repulsive core
+
+```math
+U(r) = \frac{1}{r^{12}} - \frac{2}{r^6} - \varepsilon \exp \left(-\frac{(r-r_0)^2}{2\sigma^2}\right)
```
joaander: It takes a little creativity to make U(r) have units of energy. See https://hoomd-blue.readthedocs.io/en/v5.3.0/hoomd/md/pair/ljgauss.html for how we tackled that in HOOMD-blue. We should do something similar here, or else we need to introduce new parameters that were not part of the original LJ-Gauss potential.

hoomd-interaction/src/pairwise/lennard_jones_gauss.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl IsotropicForce for LennardJonesGauss {
+    #[inline]
+    fn force(&self, r: f64) -> f64 {
+        let r_inv = r.recip();
+        let arg = -1.0 * (r - self.r_0).powi(2) / (2.0 * self.sigma_squared);
+        12.0 * (r_inv.powi(13) - r_inv.powi(7))
+            - (self.epsilon * (r - self.r_0) / self.sigma_squared) * arg.exp()
+    }
+}
```
joaander, line 149: Add a test case or two that verifies both the potential and force are computed correctly. Compare the computation to a hard-coded result value computed elsewhere.

joaander (commented):
Thanks! I have some suggestions to improve this.

joaander (approved):
Thanks!

While I don't want to include too many potentials in the main hoomd-rs repository, this one is used often enough that it is worth keeping.

PR #34, opened 2025-07-09
Implement the GSD file layer.

### LINKED ISSUE #17: Implement the GSD file layer

joaander:
The Rust implementation of the GSD file layer produces bit-for-bit identical files to the C implementation (version 4.0) when given the same sequence of write calls.

This PR implements the file layer only. Higher level helpers or abstractions for the HOOMD schema are left for later work.

Resolves #17.


joaander:
Tarpaulin reports the coverage for hoomd-gsd at: 450 of 563 (79.93%). I scrolled through the code and there are a number of false negative lines (e.g. an `Ok(())` return at the end of the function) that apparently are optimized out even in debug modes. The remaining lines that are not covered are the extensive error checks that would be time-consuming and difficult to trigger. We should not remove these checks to increase coverage.

PR #38, opened 2025-08-04
Manifolds

mthran:
The crate `hoomd-manifold` implements methods to accommodate non-Euclidean geometries, provided they have well-defined geodesic distances and have an embedding within Cartesian or Minkowski space. This pull request contains all structs and traits necessary for running monte carlo simulations on manifolds with uniform positive and negative curvature.

This PR modifies `hoomd-vector` to have the trait `Metric` which implements methods on metric spaces. The trait `Vector` is assumed to be a subtrait of `Metric`.

The hyperboloid model of hyperbolic/negatively-curved space is the upper sheet of a two-sheeted hyperboloid embedded in Minkowski space. `hoomd-manifold` implements `Hyperboloid` and `Minkowski` along with representations of SO(2,1) and SO(3,1) (via `HyperbolicAngle` and `Biquaternion`, respectively) to define methods relevant to the hyperboloid model. Local trial moves for MC are implemented in `HyperbolicTranslate`.

`Sphere` implements an embedding of an N-sphere in (N+1)-dimensional cartesian space. Local trial moves for MC are implemented in `SphericalTranslate`.


.DS_Store
joaander, line 1: Please remove all the `.DS_Store` files.

examples/Cargo.toml
```diff
@@ -27,10 +27,12 @@ hoomd-bevy = { workspace =  true, optional = true }
 hoomd-geometry.workspace = true
 hoomd-gsd.workspace = true
 hoomd-interaction.workspace = true
+hoomd-manifold.workspace = true
 hoomd-mc.workspace = true
 hoomd-microstate.workspace = true
 hoomd-utility.workspace = true
 hoomd-vector.workspace = true
+libm = "0.2.11"
```
joaander: Are we *sure* that we need libm? It is a set of fallback functions intended for use on embedded systems that lack a math library. The functions are implemented in pure Rust (e.g. https://github.com/rust-lang/compiler-builtins/blob/9c176c24e8b6295e2ba1c35d9713ef9e2d0055fb/libm/src/math/cosh.rs). That has the advantage that they can be inlined, but they may still be much slower than implementations that the compiler builtins provide (which can take advantage of architecture-specific instructions): https://www.reddit.com/r/rust/comments/y5l3zy/trigonometric_functions_in_rust_stdlib_vs_libm/

examples/Cargo.toml
```diff
@@ -56,6 +58,11 @@ required-features = ["bevy"]
 [package.metadata.example.random-walk]
 path = "mc-tutorial"
 
+[[example]]
```
joaander, line 59: As the repository is approaching the polish needed for a public release, I'd like to keep "testing" examples out of the repository. Please move these to a separate project dir outside `hoomd-rs`.

examples/LJ-hyperbolic-ratatui.rs
joaander, line 1: As the repository is approaching the polish needed for a public release, I'd like to keep "testing" examples out of the repository. Please move these to a separate project dir outside `hoomd-rs`.

hoomd-bevy/src/representation/hyperbolic_disk.rs
```diff
@@ -0,0 +1,204 @@
+// Copyright (c) 2024-2025 The Regents of the University of Michigan.
+// Part of hoomd-rs, released under the BSD 3-Clause License.
+
+/*! Implement Disk.
```
joaander:
```suggestion
/*! Implement HyperbolicDisk.
```

hoomd-bevy/src/representation/hyperbolic_disk.rs
```diff
@@ -0,0 +1,204 @@
...
+    ) where
+        I: IntoIterator<Item = (Minkowski<3>, f64)>,
+    {
+        for item in &mut query.into_iter().zip_longest(disks) {
+            match item {
+                Both((_, mut transform), (position, diameter)) => {
+                    let (poincare_position, max_projected_radius) =
+                        poincare(&position, RHO, diameter);
+                    let rad_arg =
+                        RHO * sinh(diameter / (2.0 * RHO)) / (1.0 + cosh(diameter / (2.0 * RHO)));
```
joaander: Using the std library math functions, you are welcome use `f64::sinh(arg)` instead of `arg.sinh()`.

hoomd-bevy/src/representation/hyperbolic_disk.rs
```diff
@@ -0,0 +1,204 @@
...
+    let v = diameter / (skirt * 2.0);
+    let eta = acosh(point.coordinates[2] / RHO);
+    let edge_proj = (RHO * sinh(eta - v)) / (1.0 + cosh(eta - v));
+    let rad_proj = (RHO * sinh(eta)) / (1.0 + cosh(eta)) - edge_proj;
+    ([proj[0] as f32, proj[1] as f32, 0.0_f32], rad_proj as f32)
+}
+
+/** Control how disks are rendered.
+
+[`DiskMaterial`] mixes the texture (which defaults to fully transparent) with
```
joaander:
```suggestion
[`HyperbolicDiskMaterial`] mixes the texture (which defaults to fully transparent) with
```

hoomd-interaction/src/cutoff_pair.rs
```diff
@@ -80,11 +80,11 @@ pub struct CutoffPair<E> {
...
 }
 
-impl<V, B, S, C, E> TotalEnergy<Microstate<B, S, C>> for CutoffPair<E>
+impl<M, B, S, C, E> TotalEnergy<Microstate<B, S, C>> for CutoffPair<E>
 where
     E: SitePairEnergy<S>,
-    S: Position<Vector = V>,
-    V: Vector,
+    S: Position<Metric = M>,
+    M: Metric,
```
joaander:
```suggestion
    S: Position<Point = P>,
    P: Metric,
```
I wonder if this would be more readable if we used the name `Point` (or `P` when generic) for the types that have a metric. @janbridley, @josephburkhart, any thoughts?

hoomd-microstate/Cargo.toml
```diff
@@ -14,10 +14,13 @@ version.workspace = true
 workspace = true
 
 [dependencies]
+hoomd-manifold.workspace = true
 hoomd-geometry.workspace = true
 hoomd-vector.workspace = true
 hoomd-utility.workspace = true
 rand.workspace = true
+libm = "0.2.11"
+approx = "0.5.1"
```
joaander:
```suggestion
```
approx is already listed in the dev dependencies.

hoomd-microstate/src/boundary.rs
```diff
@@ -51,7 +51,7 @@ pub(crate) const MAX_GHOSTS: usize = 8;
 // not currently possible in Rust.
 
 /** Attempt to move any body/site properties back into the simulation boundary.
-
+ *
```
joaander:
```suggestion

```
This doesn't look right in the built docs.

hoomd-microstate/src/property/point.rs
```diff
@@ -69,25 +69,155 @@ where
...
+        let transformed_point = Minkowski::from([
+            site_pos[0] * cosh(body_boost) * cos(body_theta) - site_pos[1] * sin(body_theta)
+                + site_pos[2] * sinh(body_boost) * cos(body_theta),
+            site_pos[0] * cosh(body_boost) * sin(body_theta)
+                + site_pos[1] * cos(body_theta)
+                + site_pos[2] * sinh(body_boost) * sin(body_theta),
+            site_pos[0] * sinh(body_boost) + site_pos[2] * cosh(body_boost),
+        ]);
+        let new_hyperboloid = Hyperboloid::from(&transformed_point);
+        assert_relative_eq!(
```
joaander: Is this something that needs to be checked at runtime after every trial move, even in release mode? Or do you intend this to be a debugging check? If the latter, use `debug_assert!` to validate your assumptions only in debug builds.

If it needs to be checked at runtime, you should also give a meaningful description string (https://doc.rust-lang.org/std/macro.assert_eq.html) so that users who encounter this error at run time will have some idea what they are seeing.

If you expect precise 1e-12 levels of accuracy, then you can also do a cleanup step after the trial move and project back onto the surface with *exactly* the same skirt. Without such a normalization step, the skirts will slowly drift away unless your errors are *completely* symmetric.

hoomd-microstate/src/property/point.rs
```diff
@@ -69,25 +69,155 @@ where
...
+                - site_pos[2] * sin(body_phi)
+                + site_pos[3] * sinh(body_boost) * sin(body_theta) * cos(body_phi),
+            site_pos[0] * cosh(body_boost) * sin(body_theta) * sin(body_phi)
+                + site_pos[1] * cos(body_theta) * sin(body_phi)
+                + site_pos[2] * cos(body_phi)
+                + site_pos[3] * sinh(body_boost) * sin(body_theta) * sin(body_phi),
+            site_pos[0] * sinh(body_boost) + site_pos[3] * cosh(body_boost),
+        ]);
+        let new_hyperboloid = Hyperboloid::from(&transformed_point);
+        assert_relative_eq!(
```
joaander: Same as above.

joaander (commented):
I haven't reviewed all the code yet, but here are some initial comments and questions.

examples/Cargo.toml
mthran: I removed the dependency on libm and switched to the standard f64 functions

hoomd-manifold/src/biquaternion.rs
```diff
@@ -0,0 +1,1064 @@
+// Copyright (c) 2024-2025 The Regents of the University of Michigan.
+// Part of hoomd-rs, released under the BSD 3-Clause License.
+
+/*! Implement [`Biquaternion`] anda four-dimensional matrix representation
```
janbridley:
```suggestion
/*! Implement [`Biquaternion`] and a four-dimensional matrix representation
```

hoomd-manifold/src/biquaternion.rs
```diff
@@ -0,0 +1,1064 @@
...
+
+use crate::{Error, HyperbolicRotate, HyperbolicRotationMatrix, Minkowski};
+
+/**
+## Biquaternions
+
+Biquaternions are the set of numbers $a + b\mathbf{i} + c\mathbf{j} + d\mathbf{k}$
+where $a,b,c,d$ are complex numbers and ${1,\mathbf{i},\mathbf{j},\mathbf{k}}$ are
+the quaternion algebra. Biquaternions can be thought of as a generalization of quaternions
+ which allow for complex coefficients.
```
janbridley:
```suggestion
the quaternion algebra. Biquaternions can be thought of as a generalization of quaternions
which allow for complex coefficients.
```

janbridley: It would also be nice if we restated why the biquaternions are useful - is it because they represent SO(3,1) as well?

On a side note, I'd love a little description of the differences between SO(n) and SO(n, 1) if there's an intuitive(ish) explanation of the structure/geometry of the groups

hoomd-manifold/src/biquaternion.rs
```diff
@@ -0,0 +1,1064 @@
...
+                        Complex::new(2.0,3.0),
+                        Complex::new(3.0,2.0),
+                        Complex::new(4.0,1.0)]);
+assert_eq!(4.0, q.components[0].im);
+
+```
+
+## Operations with biquaternions
+
+Similar to [`Quaternion`], biquaternions support vector operations (addition, multiplication
```
janbridley: Will need to fully qualify the path to link outside the crate I think

janbridley:
I added a few comments and notes based on a read through of the code - overall this looks excellent though, a great introduction for someone with only a minimal understanding of non-Euclidean geometry

examples/Cargo.toml
```diff
@@ -27,11 +27,13 @@ hoomd-bevy = { workspace =  true, optional = true }
 hoomd-geometry.workspace = true
 hoomd-gsd.workspace = true
 hoomd-interaction.workspace = true
+hoomd-manifold.workspace = true
 hoomd-mc.workspace = true
 hoomd-microstate.workspace = true
 hoomd-simulation.workspace = true
 hoomd-utility.workspace = true
 hoomd-vector.workspace = true
+libm = "0.2.11"
```
joaander:
```suggestion
```

hoomd-bevy/Cargo.toml
```diff
@@ -23,9 +23,11 @@ bevy.workspace = true
 bevy_diagnostic.workspace = true
 bevy_winit.workspace = true
 itertools.workspace = true
+hoomd-manifold.workspace = true
 hoomd-microstate.workspace = true
 hoomd-simulation.workspace = true
 hoomd-vector.workspace = true
+libm = "0.2.11"
```
joaander:
```suggestion
```

hoomd-bevy/src/representation/hyperbolic_disk.rs
```diff
@@ -0,0 +1,204 @@
...
+        texture::TRANSPARENT_IMAGE_HANDLE,
+    },
+    sprite::{AlphaMode2d, Material2d, Material2dPlugin},
+};
+use hoomd_manifold::{Hyperboloid, Minkowski};
+use itertools::{
+    EitherOrBoth::{Both, Left, Right},
+    Itertools,
+};
+use libm::{acosh, cosh, sinh};
```
joaander:
```suggestion
```

hoomd-bevy/src/representation/hyperbolic_disk.rs
```diff
@@ -0,0 +1,204 @@
...
+    ) where
+        I: IntoIterator<Item = (Minkowski<3>, f64)>,
+    {
+        for item in &mut query.into_iter().zip_longest(disks) {
+            match item {
+                Both((_, mut transform), (position, diameter)) => {
+                    let (poincare_position, max_projected_radius) =
+                        poincare(&position, RHO, diameter);
+                    let rad_arg =
+                        RHO * sinh(diameter / (2.0 * RHO)) / (1.0 + cosh(diameter / (2.0 * RHO)));
```
joaander:
```suggestion
                        RHO * f64::sinh(diameter / (2.0 * RHO)) / (1.0 + f64::cosh(diameter / (2.0 * RHO)));
```

hoomd-bevy/src/representation/hyperbolic_disk.rs
```diff
@@ -0,0 +1,204 @@
...
+                    let poincare_radius = (0.5)
+                        * acosh(1.0 + 2.0 * rad_arg.powi(2) / (1.0 - (rad_arg.powi(2)))) as f32;
```
joaander:
```suggestion
                        * f64::acosh(1.0 + 2.0 * rad_arg.powi(2) / (1.0 - (rad_arg.powi(2)))) as f32;
```

hoomd-bevy/src/representation/hyperbolic_disk.rs
```diff
@@ -0,0 +1,204 @@
...
+
+/// Project coordinates to Poincare disk
+/// TODO: fix radius
+fn poincare(point: &Minkowski<3>, skirt: f64, diameter: f64) -> ([f32; 3], f32) {
+    let pt = Hyperboloid::from(point);
+    let proj = pt.to_poincare();
+    let v = diameter / (skirt * 2.0);
+    let eta = acosh(point.coordinates[2] / RHO);
+    let edge_proj = (RHO * sinh(eta - v)) / (1.0 + cosh(eta - v));
+    let rad_proj = (RHO * sinh(eta)) / (1.0 + cosh(eta)) - edge_proj;
```
joaander:
```suggestion
    let eta = f64::acosh(point.coordinates[2] / RHO);
    let edge_proj = (RHO * f64::sinh(eta - v)) / (1.0 + f64::cosh(eta - v));
    let rad_proj = (RHO * f64::sinh(eta)) / (1.0 + f64::cosh(eta)) - edge_proj;
```

hoomd-geometry/src/shape/eighteight.rs
```diff
@@ -0,0 +1,154 @@
+// Copyright (c) 2024-2025 The Regents of the University of Michigan.
+// Part of hoomd-rs, released under the BSD 3-Clause License.
+
+//! Implement [`EightEight`]
+
+use crate::IsPointInside;
+use hoomd_manifold::Hyperboloid;
+use std::f64::consts::PI;
+
+/// [`EightEight`] implements a single regular octagon in the {8,8} tiling of two-dimensional hyperbolic space. The scaling of the octagon is set such that each of the angles is 2 pi/ 8, i.e., so that eight equivalent octagons may meet at each vertex.
```
joaander:
```suggestion
/// A regular octagon in two-dimensional hyperbolic space. 
///
/// [`EightEight`] implements a single regular octagon in the {8,8} tiling of
/// two-dimensional hyperbolic space. The scaling of the octagon is set such
/// that each of the angles is 2 pi/ 8 so that eight equivalent octagons
/// meet at each vertex.
```

Documentation comments start with a brief sentence defining the type/trait which is followed by a blank line and then a more complete description preferably with at least one example. Each line should be wrapped to 80 characters.

hoomd-geometry/src/shape/eighteight.rs
```diff
@@ -0,0 +1,154 @@
...
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct EightEight {
+    /// Skirt width of the hyperboloid
+    pub skirt: f64,
+}
+
+impl IsPointInside<Hyperboloid<3>> for EightEight {
+    /// Checks if a given hyperboloid point is inside [`EightEight`]
```
joaander:
```suggestion
    /// Checks if a given hyperboloid point is inside [`EightEight`].
```

hoomd-geometry/src/shape/eighteight.rs
```diff
@@ -0,0 +1,154 @@
...
+    fn is_point_inside(&self, point: &Hyperboloid<3>) -> bool {
+        EightEight::distance_to_boundary(point) >= 0.0
+    }
+}
+
+/// Cusp-to-vertex distance for {8,8} tiling for Gauss curvature K = -1
+const EIGHTEIGHT: f64 = 2.448_452_447_678_076;
+
+impl EightEight {
+    /// Computes the shortest distance between a given point and the boundary of `EightEight`. The shortest distance is along the radial path, i.e., the geodesic passing between the hyperboloid cusp and the query point.
```
joaander:
```suggestion
    /// Computes the shortest distance between a given point and the boundary of `EightEight`. 
    ///
    /// The shortest distance is along the radial path &mdash; the geodesic
    /// passing between the hyperboloid cusp and the query point.
```

hoomd-geometry/src/shape/eighteight.rs
```diff
@@ -0,0 +1,154 @@
...
+    pub fn distance_to_boundary(point: &Hyperboloid<3>) -> f64 {
+        let theta = point.point.coordinates[1].atan2(point.point.coordinates[0]);
+        let angle = theta.rem_euclid(PI / 4.0);
+        let boost = (point.point.coordinates[2] / point.skirt).acosh();
+        let tile_size = EIGHTEIGHT;
+        let eta =
+            (tile_size.tanh() / (angle.cos() - angle.sin() * (1.0 - (2.0_f64).sqrt()))).atanh();
+        point.skirt * (eta - boost)
+    }
+    /// Outputs vector of points on the boundary of the fundamental domain
```
joaander:
```suggestion
    /// Points on the boundary of the fundamental domain.
```

hoomd-geometry/src/shape/eighteight.rs
```diff
@@ -0,0 +1,154 @@
...
+    #[inline]
+    #[must_use]
+    pub fn boundary_points(m: usize, skirt: f64) -> Vec<(f64, f64)> {
```
joaander:
```suggestion
    pub fn boundary_points(m: usize, skirt: f64) -> [f64; 8] {
```
And adjust the implementation of the function accordingly. Use `std::array::from_fn` to replace the loop.

hoomd-mc/src/manifold_translate.rs
```diff
@@ -0,0 +1,250 @@
...
+use hoomd_manifold::{HyperbolicDisk, Hyperboloid, Sphere, SphericalDisk};
+use hoomd_microstate::property::Position;
+use hoomd_utility::valid::PositiveReal;
+use hoomd_vector::{Cartesian, InnerProduct};
+
+use rand::{Rng, distr::Distribution};
+
+/// Move the position of a body on a hyperbolic surface by a small distance
+///
+/// `HyperbolicTranslate` used with Sweep:
```
joaander:
```suggestion
```

hoomd-mc/src/manifold_translate.rs
```diff
@@ -0,0 +1,250 @@
...
+
+use crate::LocalTrial;
+use hoomd_manifold::{HyperbolicDisk, Hyperboloid, Sphere, SphericalDisk};
+use hoomd_microstate::property::Position;
+use hoomd_utility::valid::PositiveReal;
+use hoomd_vector::{Cartesian, InnerProduct};
+
+use rand::{Rng, distr::Distribution};
+
+/// Move the position of a body on a hyperbolic surface by a small distance
```
joaander:
```suggestion
/// Move the position of a body on a hyperbolic surface by a small distance.
```

hoomd-mc/src/manifold_translate.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// The skirt width of the hyperboloid
+    pub skirt: f64,
+}
+
+impl<B> LocalTrial<B> for HyperbolicTranslate
+where
+    B: Position<Position = Hyperboloid<3>>,
+    HyperbolicDisk: Distribution<Hyperboloid<3>>,
+{
+    /// Propose local trial moves for a body on a hyperbolic surface
```
joaander:
```suggestion
    /// Propose local trial moves for a body on a hyperbolic surface.
```

hoomd-mc/src/manifold_translate.rs
```diff
@@ -0,0 +1,250 @@
...
+        let z = (trial.position_mut().point[0].powi(2)
+            + trial.position_mut().point[1].powi(2)
+            + self.skirt.powi(2))
+        .sqrt();
+        trial.position_mut().point[2] = z;
+        trial
+    }
+}
+
+/// Move the position of a body on the surface of a sphere by a small distance
```
joaander:
```suggestion
/// Move the position of a body on the surface of a sphere by a small distance.
```

hoomd-mc/src/manifold_translate.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// );
+    ///
+    /// // Translation move does not translate the point more than a distance d away
+    /// assert!(
+    ///     d > Sphere::from(&new_body_properties.position()).distance(
+    ///         &Sphere::from(&Cartesian::from([
+    ///             2.0_f64.sqrt(),
+    ///             2.0_f64.sqrt(),
+    ///             0.0
+    ///         ]))
```
joaander:
```suggestion
    ///         &Sphere::from(&initial_point)
```

hoomd-meshless-voronoi/src/lib.rs
joaander, line 1: Remove `hoomd-meshless-voronoi`. Please make a separate pull request for meshless-voronoi when it is ready.

hoomd-microstate/Cargo.toml
```diff
@@ -14,16 +14,17 @@ version.workspace = true
 workspace = true
 
 [dependencies]
+hoomd-manifold.workspace = true
 hoomd-geometry.workspace = true
 hoomd-vector.workspace = true
 hoomd-utility.workspace = true
 rand.workspace = true
+approx = "0.5.1"
```
joaander:
```suggestion
approx.workspace = true
```

hoomd-microstate/src/property/point.rs
```diff
@@ -65,27 +66,163 @@ where
...
     }
 
     #[inline]
-    fn position_mut(&mut self) -> &mut V {
+    fn position_mut(&mut self) -> &mut P {
         &mut self.position
     }
 }
 
+impl<const N: usize> Position for Hyperboloid<N> {
```
joaander: Does `Hyperboloid<N>` need really `Position`? This should probably be removed.

The `impl<P> Position for Point<P>` above should implement `Position` correctly for `Point<Hyperboloid<N>>`.

hoomd-microstate/src/property/point.rs
```diff
@@ -65,27 +66,163 @@ where
...
+    fn position(&self) -> &Hyperboloid<N> {
+        self
+    }
+    #[inline]
+    fn position_mut(&mut self) -> &mut Hyperboloid<N> {
+        self
+    }
+}
+
+// TODO: tests.
```
joaander: Yes, please add tests for all the new `Transform` implementations.

hoomd-microstate/src/boundary/periodic/eighteight.rs
```diff
@@ -0,0 +1,659 @@
+// Copyright (c) 2024-2025 The Regents of the University of Michigan.
+// Part of hoomd-rs, released under the BSD 3-Clause License.
+
+//! Implement periodic boundary conditions for the {8,8} tiling of hyperbolic space. Specifically, `Periodic<EightEight>` identifies opposite edges of the octagon to implement the Bolza surface.
```
joaander:
```suggestion
//! Implement periodic boundary conditions for the {8,8} tiling of hyperbolic space.
//!
//! Specifically, `Periodic<EightEight>` identifies opposite edges of the octagon
//! to implement the Bolza surface.
```

hoomd-microstate/src/boundary/periodic/eighteight.rs
```diff
@@ -0,0 +1,659 @@
...
+        Error, GenerateGhosts, MAX_GHOSTS, MaximumAllowableInteractionRange, Periodic, Wrap,
+    },
+    property::Position,
+};
+use hoomd_geometry::shape::EightEight;
+use hoomd_manifold::{Hyperboloid, Minkowski};
+use hoomd_vector::Metric;
+
+impl MaximumAllowableInteractionRange for EightEight {
+    /// The largest value that the maximum interaction range can take. This bound is smaller than the one strictly imposed by the geometry by about a factor of 3.
```
joaander:
```suggestion
    /// The largest value that the maximum interaction range can take.
    ///
    /// This bound is smaller than the one strictly imposed by the geometry
    /// by about a factor of 3.
```
Why not be more exact? You seem to have a more specific implementation in the unused file.

hoomd-microstate/src/boundary/periodic/eighteight.rs
```diff
@@ -0,0 +1,659 @@
...
+    #[inline]
+    fn maximum_allowable_interaction_range(&self) -> f64 {
+        self.skirt * 1.0
+    }
+}
+
+/// Cusp-to-vertex distance for {8,8} tiling for Gauss curvature K = -1
+const EIGHTEIGHT: f64 = 2.448_452_447_678_076;
+/// Length of one of the sides of the {8,8} tiling for Gauss curvature K = -1
+const EDGE_LENGTH: f64 = 3.057_141_838_961_997;
```
joaander: Move these constants to the definition of the geometry in `hoomd-geometry` to avoid repeating them throughout the code. `EIGHTEIGHT` is used several places in both the code and examples.

You can expose a constant through a type with a const fn (rename as appropriate):
```
impl EightEight {
    const EIGHTEIGHT: f64 = 2.448_452_447_678_076;
    const EDGE_LENGTH: f64 = 3.057_141_838_961_997;
    }
...
```
and refer to the constants by the names:
```
EightEight::EIGHTEIGHT
```
and
```
EightEight::EDGE_LENGTH
```

hoomd-microstate/src/boundary/periodic/eighteight.rs
```diff
@@ -0,0 +1,659 @@
...
+
+impl<P> Wrap<P> for Periodic<EightEight>
+where
+    P: Position<Position = Hyperboloid<3>>,
+{
+    /// Wrap a point on the hyperboloid to the inside of the {8,8} tile. Note that the function fails to wrap points that are further than 0.5 + `EIGHTEIGHT` from the cusp.
```
joaander:
```suggestion
    /// Wrap a point on the hyperboloid to the inside of the {8,8} tile.
    ///
    /// Note that the function fails to wrap points that are further than 0.5 + `EIGHTEIGHT` from the cusp.
```
Explain how it fails. 
* Does it produce incorrect results? Does it return an `Err`?
* Does this limitation apply to interior points as well?

hoomd-microstate/src/boundary/eighteight.rs
joaander, line 1: This file appears to be unused. Remove it.

hoomd-mc/src/manifold_translate.rs
```diff
@@ -0,0 +1,250 @@
...
+///     microstate.increment_step();
+/// }
+/// # Ok(())
+/// # }
+/// ```
+pub struct HyperbolicTranslate {
+    /// The max distance a body can be translated in one trial move
+    pub maximum_distance: PositiveReal,
+    /// The skirt width of the hyperboloid
+    pub skirt: f64,
```
joaander: You store the skirt here *and* you implement `LocalTrial<B>` for `HyperbolicTranslate` where `B: Position<Position = Hyperboloid<3>>`. `Hyperboloid` stores the skirt as well, so why is this one necessary?

Assuming the two skirts are the same, why not use the existing `Translate` type? You can leave the implementations in this file to keep them separate, but you should be able to remove `HyperbolicTranslate` implement `LocalTrial` for `Translate` where `B: Position<Position = Hyperboloid<3>>`. You'll need to add the position type to `LocalTrial`:
```
trait LocalTrial<P, B>
```
```
impl<B> LocalTrial<Hyperboloid<3>, B> for Translate
where
    B: Position<Position = Hyperboloid<3>>,
```

hoomd-manifold/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+hoomd-vector.workspace = true
+num = "0.4.3"
+approx = "0.5.1"
+
+[dev-dependencies]
+divan.workspace = true
+paste = "1.0.15"
+rstest.workspace = true
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+ratatui = "0.28.0"
```
joaander:
```suggestion
```
Unused.

hoomd-manifold/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+keywords.workspace = true
+categories.workspace = true
+
+[dependencies]
+thiserror.workspace = true
+rand.workspace = true
+hoomd-utility.workspace = true
+hoomd-vector.workspace = true
+num = "0.4.3"
+approx = "0.5.1"
```
joaander:
```suggestion
approx.workspace = true
```

hoomd-manifold/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+    html_favicon_url = "https://hoomd-blue.readthedocs.io/en/latest/_static/hoomdblue-logo-favicon.svg"
+)]
+#![doc(
+    html_logo_url = "https://hoomd-blue.readthedocs.io/en/latest/_static/hoomdblue-logo-favicon.svg"
+)]
+
+//! Tools for non-Euclidean geometries. The trait [`CurvedManifold`] implements the `geodesic_distance` method
+//! for calculating geodesic distances for manifolds imbedded in a metric vector space. This crate Includes the struct
+//! [`Sphere`], an embedding of an N-sphere within (N+1)-dimensional cartesian space, as well as [`Hyperboloid`], the
+//! surface of the upper sheet of an N-dimensional two-sheeted hyperboloid embedded in (N+1)-dimensional Minkowski space.
```
joaander: Revise this. It references many types and methods that no longer exist.

hoomd-manifold/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+//! ## Sphere
+//!
+//! [`Sphere`] describes an N-sphere of radius R embedded in [`Cartesian<N+1>`]. By definition,
+//! the components of a point on an N-sphere satisfy
+//! ```math
+//! \sum_{i=1}^{N+1}x_i^2 = R^2
+//! ```
+//! [`Sphere`] implements a distance metric which calculates the geodesic distance on the
+//! surface of an N-sphere. Use [`Sphere`] to describe spaces with constant postive curvature.
+//!
```
joaander, line 25:
```suggestion
//! [`Cartesian<N+1>`]: hoomd_vector::Cartesian
```

hoomd-manifold/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+//! let boosted = matrix.hyperbolic_rotate(&v);
+//! // rotated is approximately [1.0,sinh(0.5),cosh(0.5)]);
+//! ```
+//!
+//! For three-dimensional hyperbolic surfaces, use [`Biquaternion`]. Biquaternions are a
+//! generalization of quaternions which allow for complex coefficients. Unit biquaternions give
+//! a representation of SO(3,1); this can either be done by converting the biquaternions
+//! to a [`HyperbolicRotationMatrix`] or by using the [`UnitBiquaternion`] algebra directly.
+//!
+//! ```math
```
joaander: This looks like an example code block, not LaTeX code. I removed `math`, but now the test fails. Please fix.

hoomd-manifold/src/biquaternion.rs
```diff
@@ -0,0 +1,1133 @@
...
+    array, fmt,
+    iter::zip,
+    ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Sub, SubAssign},
+};
+
+use crate::{Error, HyperbolicRotate, HyperbolicRotationMatrix, Minkowski};
+#[expect(unused_imports, reason = "Needed for doc link")]
+use hoomd_vector::Quaternion;
+
+/// ## Biquaternions
```
joaander:
```suggestion
/// A quaternion with complex coefficients.
```

hoomd-manifold/src/biquaternion.rs
joaander, line 1: There are many issues to fix in the documentation for biquaternion. Please fix these common issues throughout the whole file. I'm not going to comment on each of them individually.
* Method descriptions are not sentences.
* Many math equations do not render.

hoomd-manifold/src/biquaternion.rs
```diff
@@ -0,0 +1,1133 @@
...
+    #[inline]
+    fn div_assign(&mut self, rhs: f64) {
+        for n in 0..4 {
+            self.components[n] /= Complex::new(rhs, 0.0);
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+/// ## Representation of SO(3,1)
```
joaander:
```suggestion
/// Represent SO(3,1) with a normalized biquaternion.
```

hoomd-manifold/src/hyperbolic_angle.rs
```diff
@@ -0,0 +1,349 @@
...
+use rand::{
+    Rng,
+    distr::{Distribution, StandardUniform, Uniform},
+};
+use std::{f64::consts::PI, fmt};
+
+use crate::HyperbolicRotationMatrix;
+
+/// ## Hyperbolic rotations
+/// "Hyperbolic rotations" describe elements of the group SO(N,1), which preserve hyperboloids
```
joaander:
```suggestion
/// Combine rotation and boosts in Hyperbolic space.
///
/// "Hyperbolic rotations" describe elements of the group SO(N,1), which preserve hyperboloids
```
Or a more appropriate definition formatted as a sentence.

hoomd-manifold/src/hyperbolic_angle.rs
```diff
@@ -0,0 +1,349 @@
...
+/// embedded in [`Minkowski<N+1>`]. Transformations include pure spatial rotations as well as
+/// "boosts".
+///
```
joaander, line 18:
```suggestion
///
/// [`Minkowski<N+1>`]: Minkowski
```
Check that the link is formed correctly after making this change.

hoomd-manifold/src/hyperbolic_angle.rs
```diff
@@ -0,0 +1,349 @@
...
+/// For two-dimensional hyperbolic surfaces, use [`HyperbolicAngle`] to implement
+/// elements of SO(2,1) which rotate points about the z-axis or boost points along the x- and y-axes.
+/// SO(2,1) transformations represent symmetries of two-dimensional hyperbolic space. The algebra
+/// so(2,1) is generated by the matrices
+/// ```math
+/// K_x = \begin{bmatrix} 0&0&1\\ 0&0&0\\1&0&0 \end{bmatrix} \qquad\qquad
+/// K_y = \begin{bmatrix} 0&0&0\\ 0&0&1\\0&1&0 \end{bmatrix}\qquad\qquad
+/// J_z = \begin{bmatrix} 0&-1&0\\ 1&0&0\\0&0&0 \end{bmatrix}
+/// ``` which satisfy commutation relations
```
joaander:
```suggestion
/// ```
/// which satisfy commutation relations
```

hoomd-manifold/src/hyperbolic_angle.rs
```diff
@@ -0,0 +1,349 @@
...
+/// ```math
+/// G = \exp(aJ_z + b K_x + c K_y)
+/// ```
+/// In particular,
+/// ```math
+/// \exp(aJ_z) = \begin{bmatrix}\cos(a)&-\sin(a)&0\\ \sin(a)&\cos(a)&0 \\ 0&0&1\end{bmatrix}
+/// \\ \exp(bK_x) = \begin{bmatrix}\cosh(b)&0&\sinh(b)\\ 0&1&0 \\ \sinh(b)&0&\cosh(b)\end{bmatrix}
+/// \\ \exp(cK_y) = \begin{bmatrix}1&0&0\\ 0&\cosh(c)&\sinh(c) \\ 0&\sinh(c)&\cosh(c)\end{bmatrix}
+/// ```
+/// [`HyperbolicAngle`] is a struct of three real numbers $(a,b,c)$ which parameterize the generators
```
joaander:
```suggestion
/// [`HyperbolicAngle`] is a struct of three real numbers $`(a,b,c)`$ which parameterize the generators
```

hoomd-manifold/src/hyperbolic_angle.rs
```diff
@@ -0,0 +1,349 @@
...
+    }
+}
+
+impl Distribution<HyperbolicAngle> for StandardUniform {
+    /// Sample a random hyperbolic angle which generates an (approximately) uniform
+    /// distribution with maximum boost distance $v=1$ over the two-dimensional hyperboloid
+    /// with skirt width $rho = 1$.
+    ///
+    /// Every point on the hyperboloid can be generated by a boost in the x-direction followed
+    /// by a rotation. The Baker-Campbell-Hausdorff Lemma (<https://en.wikipedia.org/wiki/Baker%E2%80%93Campbell%E2%80%93Hausdorff_formula>)
```
joaander:
```suggestion
    /// by a rotation. The [Baker-Campbell-Hausdorff Lemma](<https://en.wikipedia.org/wiki/Baker%E2%80%93Campbell%E2%80%93Hausdorff_formula>)
```

hoomd-manifold/src/hyperbolic_angle.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// gives an expression for the relationship between subsequent SO(2,1) transformations
+    /// and the so(2,1) algebra generators. Explicitly,
+    ///
+    /// ```math
+    /// \exp[\theta J_z]\exp[v K_x] = \exp[(\theta + \frac{1}{12}\theta v^2)J_z + (v-\frac{1}{12}\theta^2v)K_x + \frac{1}{2}\theta v K_y + \text{higher order}]
+    /// ```
+    /// To avoid clustering around the cusp of the hyperboloid, the random boost parameter $v$ is
+    /// sampled from a uniform distribution between $0$ and $1$ and then passed through a square
+    /// root function.
```
joaander: Does this make the distribution uniform? Or just less clustered?

The distribution doesn't *have* to be uniform, but you should clearly document exactly *what* distribution it does sample from.

hoomd-manifold/src/hyperbolic_angle.rs
```diff
@@ -0,0 +1,349 @@
...
+///     (0.5_f64).sinh(),
+///     epsilon = 1e-12
+/// );
+/// assert_relative_eq!(
+///     boosted.coordinates[2],
+///     (0.5_f64).cosh(),
+///     epsilon = 1e-12
+/// );
+/// ```
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
```
joaander, line 93: `#[derive(Default)`] will make `HyperbolicAngle::Default` return `(0.0, 0.0, 0.0)`. Is that appropriate? Explain what the default value is in the type documentation.

hoomd-manifold/src/hyperbolic_angle.rs
joaander, line 1: I've commented specifically on some lines that need to be fixed in the documentation. In addition, please review the whole generated documentation page. There are more places where equations don't render, links don't link, or need general revisions to the wording.

hoomd-manifold/src/minkowski.rs
```diff
@@ -0,0 +1,1256 @@
...
+    f64::consts::PI,
+    fmt,
+    iter::zip,
+    ops::{Add, AddAssign, Div, DivAssign, Index, IndexMut, Mul, MulAssign, Neg, Sub, SubAssign},
+};
+
+#[cfg(debug_assertions)]
+use approx::assert_relative_eq;
+
+/// [`Minkowski<N>`] implements (N-1,1)-dimensional Minkowski space with the metric signature
```
joaander: Documentation blocks should start with a short single sentence on one line, followed by a blank line, then a more extended description.

hoomd-manifold/src/minkowski.rs
joaander, line 1: There are many equations that don't render and links that don't link in this file. I'm not going to individually comment on all of them.

hoomd-manifold/src/minkowski.rs
```diff
@@ -0,0 +1,1256 @@
...
+/// ```
+///
+/// The distance metric on Minkowski space is given by the "spacetime interval"
+/// ```math
+/// d_M^2(\vec{u},\vec{v}) = (\vec{u}-\vec{v})^T \eta (\vec{u}-\vec{v})
+/// = (u_1-v_1)^2 +\cdots + (u_{N-1}-v_{N-1})^2 - (u_N - v_N)^2
+/// ```
+/// Note that because this metric is not positive-definite, [`Minkowski`] only implements a
+/// squared distance metric (i.e., it does not implement a "distance" function which
+/// takes the square root of `distance_squared`).
```
joaander: But there is an implementation of the `distance` method. Perhaps we should break `Metric` up into two traits so that `Minkowski` can in reality only implement `MetricSquared` and not `Metric`.

hoomd-manifold/src/minkowski.rs
```diff
@@ -0,0 +1,1256 @@
...
+        )]
+        let uniform = Uniform::new_inclusive(-1.0, 1.0)
+            .expect("hard-coded range should form a valid distribution");
+        Minkowski {
+            coordinates: array::from_fn(|_| uniform.sample(rng)),
+        }
+    }
+}
+
+/// ## Hyperboloid Model
```
joaander:
```suggestion
/// A surface with constant negative curvature.
///
```

hoomd-manifold/src/minkowski.rs
```diff
@@ -0,0 +1,1256 @@
...
+/// The trait [`Hyperboloid`] implements an embedding of the top sheet of an (N-1)-dimensional
```
joaander:
```suggestion
/// [`Hyperboloid`] implements an embedding of the top sheet of an (N-1)-dimensional
```
Hyperboloid is a type.

hoomd-manifold/src/minkowski.rs
```diff
@@ -0,0 +1,1256 @@
...
+/// ```math
+/// x_1^2 +\cdots x_{N-1}^2 - x_{N}^2 = -R^2
+/// ```
+/// Where the "top sheet" is defined by the $`x_N>0`$ solutions. In Minkowski space, the hyperboloid
+/// has a natural interpretation as the set of points with the same spacetime interval
+/// ```math
+/// \Delta s^2 = \vec{x}^T \eta \vec{x} = x_1^2 +\cdots x_{N-1}^2 - x_{N}^2
+/// ```
+///
+/// [`Hyperboloid`] defines a distance metric [`hyperbolic_distance`] which computes the distance
```
joaander: The method `hyperbolic_distance` does not exist. Please update the documentation to reflect the current implementation.

hoomd-manifold/src/minkowski.rs
```diff
@@ -0,0 +1,1256 @@
...
+    /// Get the skirt width of the hyperboloid
+    #[must_use]
+    #[inline]
+    pub fn skirt(&self) -> f64 {
+        self.skirt
+    }
+    /// Create a hyperboloid point from a Minkowski vector
+    #[must_use]
+    #[inline]
+    pub fn from(point: &Minkowski<N>) -> Hyperboloid<N> {
```
joaander:
```suggestion
    pub fn from(point: Minkowski<N>) -> Hyperboloid<N> {
```
`from` should take ownership of its input.

hoomd-manifold/src/minkowski.rs
```diff
@@ -0,0 +1,1256 @@
...
+        let skirt_squared = -point.distance_squared(&Minkowski::<N>::default());
+        Hyperboloid {
+            point: *point,
+            skirt: skirt_squared.sqrt(),
+        }
```
joaander: If skirt can be computed given the vector in Minkowski space, they why do you allow callers to set it directly as a public field? Why do you even store the skirt at all? Does it get used in many places so that it needs to be precomputed and cached?

hoomd-manifold/src/minkowski.rs
```diff
@@ -0,0 +1,1256 @@
...
+        let phi_mod = phi.rem_euclid(PI);
+        let point = Minkowski::from([
+            skirt * (v.sinh()) * (theta_mod.cos()),
+            skirt * (v.sinh()) * (theta_mod.sin()) * (phi_mod.cos()),
+            skirt * (v.sinh()) * (theta_mod.sin()) * (phi_mod.sin()),
+            skirt * (v.cosh()),
+        ]);
+        Hyperboloid::from(&point)
+    }
+    /// Computes the length of the geodesic passing between the cusp $(0,\cdots,0,\rho)$ and a given
```
joaander: Shorter definition needed before the full description.

hoomd-manifold/src/minkowski.rs
```diff
@@ -0,0 +1,1256 @@
...
+        }
+    }
+}
+
+/// [`Metric`] implements the metric for the hyperboloid model of hyperbolic space
+impl Metric for Hyperboloid<3> {
+    #[inline]
+    fn distance(&self, other: &Self) -> f64 {
+        #[cfg(debug_assertions)]
+        assert_relative_eq!(self.skirt, other.skirt, epsilon = 1e-12);
```
joaander: Am I correct to assume that this distance metric is only valid for two points on the same hyperboloid (with the same skirt width)?

If so, then I would check this in both release and debug builds to ensure that users' programs throw a reasonable error instead of returning a nonsensical result. At this time, I don't know whether to suggest an absolute or relative check. At least provide a reason so that users will see a meaningful error message.
```suggestion
        assert!(relative_eq!(self.skirt, other.skirt, epsilon = 1e-12), "points must be on the same hyperboloid");
```

hoomd-manifold/src/minkowski.rs
```diff
@@ -0,0 +1,1256 @@
...
+            *result = zip(row.coordinates.iter(), vector.coordinates.iter())
+                .fold(0.0, |product, x| product + x.0 * x.1);
+        }
+
+        Minkowski { coordinates }
+    }
+}
+
+/// A uniform distribution of points within distance r of a point on the 2-dimensional hyperboloid
+/// with a given skirt width.
```
joaander: Needs a shorter definition, blank line, and then the full description.

Since we are talking about curved space here, you should define *what* is uniform about the distribution.

hoomd-manifold/src/sphere.rs
```diff
@@ -0,0 +1,314 @@
...
+    distr::{Distribution, Uniform},
+};
+
+use hoomd_utility::valid::PositiveReal;
+use hoomd_vector::{Cartesian, InnerProduct, Metric};
+
+#[cfg(debug_assertions)]
+use approx::assert_relative_eq;
+
+/// The trait [`Sphere`] for [`Cartesian`] implements types on the embedding of an N-sphere in Euclidean space.
```
joaander: It is a type, not a trait. This also needs to be formatted as a short definition sentence followed by a blank line, followed by the full description.

hoomd-manifold/src/sphere.rs
```diff
@@ -0,0 +1,314 @@
...
+    /// Get the radius of the sphere
+    #[inline]
+    #[must_use]
+    pub fn radius(&self) -> f64 {
+        self.radius
+    }
+    /// Create a sphere point from a cartesian vector
+    #[inline]
+    #[must_use]
+    pub fn from(point: &Cartesian<N>) -> Sphere<N> {
```
joaander:
```suggestion
    pub fn from(point: Cartesian<N>) -> Sphere<N> {
```
from should take ownership.

hoomd-manifold/src/sphere.rs
```diff
@@ -0,0 +1,314 @@
...
+        let rad = point.norm();
+        Sphere {
+            point: *point,
+            radius: rad,
+        }
+    }
+    /// Create a 2-sphere from spherical coordinates
+    #[inline]
+    #[must_use]
+    pub fn from_2_angles(r: f64, theta: f64, phi: f64) -> Sphere<3> {
```
joaander: It seems weird to me that there is a method in the `impl Sphere<N>` block that produces `Sphere<3>`. `Sphere<1000>::from_2_angles` giving a `Sphere<3>` does not make sense.

I would move this function to an `impl Sphere<3>` block and name it `from_polar_coordinates`. Do the same for `from_3_angles`. Then the user can call
```
Sphere<3>::from_polar_coordinates
```
or
```
Sphere<4>::from_polar_coordinates
```
and obtain what they expect.

hoomd-manifold/src/sphere.rs
```diff
@@ -0,0 +1,314 @@
...
+            .collect::<Vec<f64>>()
+    }
+}
+
+/// [`Metric`] implements the positivvely curved metric on the surface of a sphere
+impl Metric for Sphere<3> {
+    #[inline]
+    fn distance(&self, other: &Self) -> f64 {
+        #[cfg(debug_assertions)]
+        assert_relative_eq!(self.radius, other.radius, epsilon = 1e-12);
```
joaander: This should be checked at runtime for the same reason discussed in Hyperboloid.

hoomd-manifold/src/sphere.rs
```diff
@@ -0,0 +1,314 @@
...
+    /// Max distance away from point
+    pub r: PositiveReal,
+    /// The center of the disk
+    pub point: Cartesian<3>,
+    /// The radius of the sphere
+    pub radius: f64,
+}
+
+impl Distribution<Sphere<3>> for SphericalDisk {
+    /// Translates 3-dimensional cartesian vector named "point" along the surface of a sphere by maximum distance of r.
```
joaander: The same doc formatting comments given for HyperbolicDisk apply here.

hoomd-manifold/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+
+#![doc(
+    html_favicon_url = "https://hoomd-blue.readthedocs.io/en/latest/_static/hoomdblue-logo-favicon.svg"
+)]
+#![doc(
+    html_logo_url = "https://hoomd-blue.readthedocs.io/en/latest/_static/hoomdblue-logo-favicon.svg"
+)]
+
+//! Tools for non-Euclidean geometries. The trait [`CurvedManifold`] implements the `geodesic_distance` method
+//! for calculating geodesic distances for manifolds imbedded in a metric vector space. This crate Includes the struct
```
joaander:
```suggestion
//! for calculating geodesic distances for manifolds embedded in a metric vector space. This crate Includes the struct
```
Please install and use a code spell checker. There are other errors that I didn't fix. I use [codebook](https://github.com/blopker/codebook) and include a project-specific dictionary in the repository. It doesn't yet have a published vscode extension, but hopefully will soon. [cspell](https://streetsidesoftware.com/vscode-spell-checker/) also works well and does have a vscode extension.

joaander (changes requested):
Thanks! I have a lot of comments to improve the documentation and some stylistic improvements (function names, placement, etc...).

I think there is one big design question that we need to resolve around the replication of `skirt` and `radius`, whether they should be stored at all, where and how the caller should be allowed to set them, and who is responsible for validating them. I think I see where you are going with this design, but we should talk offline about it. The big problem with it currently is that the caller can do this:
```
let v = Sphere {
    point: Cartesian::from([5.0, 0.0, 0.0]),
    radius: 1.0,
}
```

hoomd-microstate/src/property/point.rs
```diff
@@ -65,32 +63,157 @@ where
...
             position: self.position + site_properties.position,
         }
     }
 }
 
-impl<V> Position for Point<V> {
-    type Vector = V;
+impl Transform<Point<Hyperboloid<3>>> for Point<Hyperboloid<3>> {
+    #[inline]
+    fn transform(&self, site_properties: &Point<Hyperboloid<3>>) -> Point<Hyperboloid<3>> {
```
joaander: Document this method. Describe how sites in a body will be transformed.

hoomd-microstate/src/property/point.rs
```diff
@@ -65,32 +63,157 @@ where
...
+            site_pos[0] * (body_boost.sinh()) + site_pos[2] * (body_boost.cosh()),
+        ]);
+        let new_hyperboloid = Hyperboloid::from(transformed_point, skirt);
+        Point::new(new_hyperboloid)
+    }
+}
+
+impl Transform<Point<Hyperboloid<4>>> for Point<Hyperboloid<4>> {
+    #[inline]
+    fn transform(&self, site_properties: &Point<Hyperboloid<4>>) -> Point<Hyperboloid<4>> {
```
joaander: Document this method. Describe how sites in a body will be transformed.

hoomd-microstate/src/property/point.rs
```diff
@@ -65,32 +63,157 @@ where
...
+            site_pos[0] * (body_boost.sinh()) + site_pos[3] * (body_boost.cosh()),
+        ]);
+        let new_hyperboloid = Hyperboloid::from(transformed_point, skirt);
+        Point::new(new_hyperboloid)
+    }
+}
+
+impl Transform<Point<Sphere<3>>> for Point<Sphere<3>> {
+    #[inline]
+    fn transform(&self, site_properties: &Point<Sphere<3>>) -> Point<Sphere<3>> {
```
joaander: Document this method. Describe how sites in a body will be transformed.

hoomd-microstate/src/property/point.rs
```diff
@@ -65,32 +63,157 @@ where
...
+            -trial_coords[0] * (body_theta.sin()) + trial_coords[2] * (body_theta.cos()),
+        ]);
+        let new_sphere = Sphere::from(transformed_point, radius);
+        Point::new(new_sphere)
+    }
+}
+
+impl Transform<Point<Sphere<4>>> for Point<Sphere<4>> {
+    #[inline]
+    fn transform(&self, site_properties: &Point<Sphere<4>>) -> Point<Sphere<4>> {
```
joaander: Document this method. Describe how sites in a body will be transformed.

hoomd-microstate/src/property/point.rs
joaander: I see that commenting this out causes compile errors, but I don't understand why. Need to look into this further.

hoomd-manifold/src/minkowski.rs
```diff
@@ -0,0 +1,1283 @@
...
+            .map(|i| {
+                self.point.coordinates[*i] / (1.0 + self.point.coordinates[N - 1] / self.skirt)
+            })
+            .collect::<Vec<f64>>()
+    }
+}
+
+impl<const N: usize> Default for Hyperboloid<N> {
+    #[inline]
+    fn default() -> Self {
```
joaander, line 641: Document what skirt with is selected and what point is the "default".

hoomd-manifold/src/minkowski.rs
```diff
@@ -0,0 +1,1283 @@
...
+            point: zero,
+            skirt: 1.0_f64,
+        }
+    }
+}
+
+/// [`Metric`] implements the metric for the hyperboloid model of hyperbolic space
+impl Metric for Hyperboloid<3> {
+    #[inline]
+    fn distance(&self, other: &Self) -> f64 {
```
joaander, line 663: Add documentation.

hoomd-manifold/src/minkowski.rs
```diff
@@ -0,0 +1,1283 @@
...
+    }
+    #[inline]
+    fn n_dimensions(&self) -> usize {
+        2_usize
+    }
+}
+
+impl Metric for Hyperboloid<4> {
+    #[inline]
+    fn distance(&self, other: &Self) -> f64 {
```
joaander, line 700: Add documentation.

hoomd-manifold/src/sphere.rs
```diff
@@ -0,0 +1,349 @@
...
+            r * (theta_mod.cos()),
+        ]);
+        Sphere::from_cartesian_coordinates(point, r)
+    }
+}
+
+/// [`Metric`] implements the positively curved metric on the surface of a sphere
+impl Metric for Sphere<3> {
+    #[inline]
+    fn distance(&self, other: &Self) -> f64 {
```
joaander, line 143: Add documentation.

hoomd-manifold/src/sphere.rs
```diff
@@ -0,0 +1,349 @@
...
+    }
+    #[inline]
+    fn n_dimensions(&self) -> usize {
+        2_usize
+    }
+}
+
+impl Metric for Sphere<4> {
+    #[inline]
+    fn distance(&self, other: &Self) -> f64 {
```
joaander, line 174: Add Documentation.

hoomd-manifold/src/sphere.rs
```diff
@@ -0,0 +1,349 @@
...
+/// let transformed_random_point: Sphere<3> = disk.sample(&mut rng);
+///
+/// assert!(0.1 > random_point.distance(&transformed_random_point));
+///
+/// # Ok(())
+/// # }
+/// ```
+pub struct SphericalDisk {
+    /// Max distance away from point
+    pub r: PositiveReal,
```
joaander: Use a better name for `r`. It is easy to confuse with `radius`.

hoomd-manifold/src/sphere.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// The center of the disk
+    pub point: Cartesian<3>,
+    /// The radius of the sphere
+    pub radius: f64,
```
joaander: I think this should store a `Sphere` directly as a field, rather than the fields of `Sphere`.
```suggestion
    pub point: Sphere<3>,
```

hoomd-manifold/src/minkowski.rs
```diff
@@ -0,0 +1,1283 @@
...
+/// # Ok(())
+/// # }
+/// ```
+pub struct HyperbolicDisk {
+    /// Max distance away from point
+    pub r: PositiveReal,
+    /// The center of the disk
+    pub point: Minkowski<3>,
+    /// The skirt width of the hyperboloid
+    pub skirt: f64,
```
joaander:
```suggestion
    pub point: Hyperbolic<3>,
```
I think this should store Hyperbolic directly rather than its fields.

hoomd-mc/src/manifold_translate.rs
```diff
@@ -0,0 +1,152 @@
...
+            radius: trial.position().radius(),
+        };
+        *trial.position_mut() = disk.sample(rng);
+        let rescale = trial.position().radius() / trial.position().point().norm();
+        *trial.position_mut() = Sphere::from_cartesian_coordinates(
+            *trial.position().point() * rescale,
+            trial.position().radius());
+        trial
+    }
+}
```
joaander: Add more extensive unit tests.

hoomd-mc/src/manifold_translate.rs
```diff
@@ -0,0 +1,152 @@
...
+        let disk = SphericalDisk {
+            r: self.maximum_distance,
+            point: *trial.position_mut().point(),
+            radius: trial.position().radius(),
+        };
+        *trial.position_mut() = disk.sample(rng);
+        let rescale = trial.position().radius() / trial.position().point().norm();
+        *trial.position_mut() = Sphere::from_cartesian_coordinates(
+            *trial.position().point() * rescale,
+            trial.position().radius());
```
joaander: While this adds safety, I don't think it is necessary. The way `disk.sample` is constructed ensures that the generated trial is always on the surface of the sphere.

If there is indeed drift without this rescaling, then I would add a new method to `Sphere`: `project_cartesian` that does the rescaling on initialization. That would be generally useful in a number of circumstances.

hoomd-manifold/src/minkowski.rs
```diff
@@ -0,0 +1,1313 @@
...
+            trial_coords[0] * (eta.cosh()) * (phi.cos()) - trial_coords[1] * (phi.sin())
+                + trial_coords[2] * (eta.sinh()) * (phi.cos()),
+            trial_coords[0] * (eta.cosh()) * (phi.sin())
+                + trial_coords[1] * (phi.cos())
+                + trial_coords[2] * (eta.sinh()) * (phi.sin()),
+            trial_coords[0] * (eta.sinh()) + trial_coords[2] * (eta.cosh()),
+        ]);
+        let new_hyperboloid = Hyperboloid::from_minkowski_coordinates(transformed_point, rho);
+        #[cfg(debug_assertions)]
+        assert_relative_eq!(rho, new_hyperboloid.skirt(), epsilon = 1e-12);
```
joaander:
```suggestion
```
`from_minkowski_coordinates` assigns the given skirt, so this assert will always pass. It also checks that `transformed_point` is on the hyperboloid, so an additional check here is not needed.

hoomd-manifold/src/sphere.rs
```diff
@@ -0,0 +1,349 @@
...
+            trial_coords[0] * (theta.cos()) * (phi.cos()) - trial_coords[1] * (phi.sin())
+                + trial_coords[2] * (theta.sin()) * (phi.cos()),
+            trial_coords[0] * (theta.cos()) * (phi.sin())
+                + trial_coords[1] * (phi.cos())
+                + trial_coords[2] * (theta.sin()) * (phi.sin()),
+            -trial_coords[0] * (theta.sin()) + trial_coords[2] * (theta.cos()),
+        ]);
+        let new_sphere = Sphere::from_cartesian_coordinates(transformed_point, radius);
+        #[cfg(debug_assertions)]
+        assert_relative_eq!(radius, new_sphere.radius, epsilon = 1e-12);
```
joaander:
```suggestion
```
`from_cartesian_coordinates` simply assigns the radius, so this assert will always pass. It also asserts that the point is on the sphere, so this duplicate check can be removed.

hoomd-manifold/src/biquaternion.rs
```diff
@@ -0,0 +1,1104 @@
...
+            return Err(Error::InvalidBiquaternionMagnitude);
+        }
+        Ok(UnitBiquaternion(self / mag))
+    }
+    /// Create a [`UnitBiquaternion`] by normalizing the given biquaternion
+    /// without returning an Option type.
+    #[inline]
+    #[must_use]
+    pub fn to_unit_unchecked(self) -> UnitBiquaternion {
+        UnitBiquaternion(self)
```
joaander: This doesn't divide by the magnitude. Is this intentional?

Add a unit test for `to_unit_unchecked` to ensure that it operates correctly.

joaander (changes requested):
Thanks for all your work on this! I've made a full pass on the branch and made many commits to directly address my suggestions.
I revised some documentation, moved the new `P` generic from `LocalTrial` to `Translate`, mirrored the same structure in `Rotate`, and implemented `approxim` checks for the new vector types.

I have a few remaining suggestions for you to implement (see line comments below).

Bigger picture, there are a few things to discuss and possibly act on before we merge:
* What are you future plans for 'hyperbolic-lennard-jones.rs'? Do you plan to write a tutorial for this example like the others? If not, should it continue to live as a standalone code-only example for users?
* I think `Sphere` and `Hyperboloid` need better names. These are objects that represent a *point on a sphere* and a *point on a hyperboloid* respectively. Looking at just the name `Sphere`, a user could easily confuse it with the representation of a sphere in e.g. hoomd-geometry or hoomd-bevy.

hoomd-microstate/src/property/point.rs
joaander: I remove it and adjusted the other code as needed. This might break your existing production scripts. You need to use `Point<Hyperboloid>` as the Body and Site properties.

joaander (approved):
Thanks! I'm going to merge this.

I'd still like to see some more extensive tests for the local trial moves, but those can be done in a later PR: #108, #109.

PR #67, opened 2025-08-28
Add 2D camera controls to the `hoomd_bevy` crate

josephburkhart:
I've created a new resource and system for controlling a camera with an orthographic projection in 2D space. I have then added these assets to the tutorials.

Currently, the new resource and system are provided as a public struct and function, and have to be manually added to a new Bevy app. If we want to make this process automatic, we'll need to have a way to specify to the Hoomd Bevy plugin constructor that the app is 2D, since these camera controls only work in 2D.

Feedback/suggestions welcome!

### DIFF
--- a/.codebook.toml
+++ b/.codebook.toml
@@ -31,6 +31,7 @@ words = [
     "tetronimo",
     "tetronimoes",
     "timestep",
+    "trackpad",
     "versor",
     "webgl",
     "webgpu",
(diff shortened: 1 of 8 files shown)


joaander (commented):
Thanks!

I think it should be possible to use the `Camera2d` projection (in combination with the window geometry) to compute what 1 pixel is equal to in simulation units. That would remove the need for an empirical `translation_scale_factor` and make it so that clicking and dragging a location in space will keep that location under the mouse cursor.

Many games (and google maps too) do the same when scaling. The translation is updated with the scale so that the location under the mouse cursor remains fixed while scaling.

PR #69, opened 2025-09-10
Add a crate for small matrix linear algebra

janbridley:
Simple linear algebra traits, types, and subroutines. I've kept the interface for these structs as simple as possible, and have tried to balance generality with performance. I left `MatMul` and `Invertible` as traits as I figured that could be useful (and we still get compile-time size checking for matrix multiply), but I moved SVD and Determinant into inherent implementations for clarity.

I've also added benchmarks and documentation.

## TODO:
- [x] det -> determinant
- [x] Faster det for 3x3 and 4x4 cases, increase block size for recursive helper
- [x] 3x3 and 4x4 inverse
- [x] MulAssign scalar
- [x] AddAssign matrix
- [x] NegAssign matrix
- [x] More tests 
- [x] Doc examples
- [x] Left multiply by scalar
- [x] svd3
- [x] Refactor to cut down on matrix.rs file length


hoomd-linear-algebra/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+homepage.workspace = true
+keywords.workspace = true
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+version.workspace = true
+
+# [dependencies]
+# hoomd-vector.workspace = true
+
```
joaander:
```suggestion
```

hoomd-linear-algebra/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+[lints]
+workspace = true
+
+[features]
+default = []
+
```
joaander:
```suggestion
```
No need to restate the default.

hoomd-linear-algebra/benches/linalg.rs
```diff
@@ -0,0 +1,115 @@
...
+
+//! Benchmark Matrix
+
+use divan::{self, Bencher, black_box, counter::ItemsCount};
+use rand::{Rng, SeedableRng, rngs::StdRng};
+
+use hoomd_linear_algebra::{Invertible, MatMul, matrix::Matrix};
+
+fn main() {
+    #[cfg(not(target_arch = "wasm32"))]
```
joaander:
```suggestion
```
I recently refactored the CI infrastructure so that these are no longer necessary in benches.

hoomd-linear-algebra/benches/linalg.rs
```diff
@@ -0,0 +1,115 @@
...
+fn create_random_matrix_pair<const N: usize, R: Rng>(rng: &mut R) -> (Matrix<N, N>, Matrix<N, N>) {
+    (create_random_matrix(rng), create_random_matrix(rng))
+}
+
+/// Dimensions for general square matrix benchmarks.
+const SQUARE_DIMENSIONS: &[usize] = &[2, 3, 4, 8, 16, 64];
+/// Dimensions for determinant benchmarks, which are O(n!)
+const DETERMINANT_DIMENSIONS: &[usize] = &[2, 3, 4, 5, 6, 7, 8];
+
+#[cfg(not(target_arch = "wasm32"))]
```
joaander:
```suggestion
```

hoomd-linear-algebra/benches/linalg.rs
```diff
@@ -0,0 +1,115 @@
...
+fn matmul_matn<const N: usize>(bencher: Bencher) {
+    let mut rng = StdRng::seed_from_u64(42);
+
+    bencher
+        .counter(ItemsCount::from(1_u32))
+        .with_inputs(|| create_random_matrix_pair::<N, _>(&mut rng))
+        .bench_local_values(|(a, b)| black_box(a.matmul(&b)));
+}
+
+#[cfg(not(target_arch = "wasm32"))]
```
joaander:
```suggestion
```

hoomd-linear-algebra/benches/linalg.rs
```diff
@@ -0,0 +1,115 @@
...
+fn det_matn<const N: usize>(bencher: Bencher) {
+    let mut rng = StdRng::seed_from_u64(42);
+
+    bencher
+        .counter(ItemsCount::from(1_u32))
+        .with_inputs(|| create_random_matrix::<N, N, _>(&mut rng))
+        .bench_local_values(|a| black_box(a.det()));
+}
+
+#[cfg(not(target_arch = "wasm32"))]
```
joaander:
```suggestion
```

hoomd-linear-algebra/benches/linalg.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+    let mut rng = StdRng::seed_from_u64(42);
+
+    bencher
+        .counter(ItemsCount::from(1_u32))
+        .with_inputs(|| create_random_matrix::<3, 3, _>(&mut rng))
+        .bench_local_values(|a| black_box(det33(&a)));
+}
+
+#[cfg(not(target_arch = "wasm32"))]
```
joaander:
```suggestion
```

hoomd-linear-algebra/benches/linalg.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn det33(mat: &Matrix<3, 3>) -> f64 {
+        let [[a, b, c], [d, e, f], [g, h, i]] = mat.rows;
+        a * (e * i - f * h) - b * (d * i - f * g) + c * (d * h - e * g)
+    }
+    let mut rng = StdRng::seed_from_u64(42);
+
+    bencher
+        .counter(ItemsCount::from(1_u32))
+        .with_inputs(|| create_random_matrix::<3, 3, _>(&mut rng))
+        .bench_local_values(|a| black_box(det33(&a)));
```
joaander, line 70: Why do `det_mat3_fast` and `det_mat4_fast` benchmark functions that are only implemented in the benchmark?

hoomd-linear-algebra/benches/linalg.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+    let mut rng = StdRng::seed_from_u64(42);
+
+    bencher
+        .counter(ItemsCount::from(1_u32))
+        .with_inputs(|| create_random_matrix::<4, 4, _>(&mut rng))
+        .bench_local_values(|a| black_box(det44(&a)));
+}
+
+#[cfg(not(target_arch = "wasm32"))]
```
joaander:
```suggestion
```

hoomd-linear-algebra/benches/linalg.rs
```diff
@@ -0,0 +1,115 @@
...
+fn inverse_mat2(bencher: Bencher) {
+    let mut rng = StdRng::seed_from_u64(42);
+
+    bencher
+        .counter(ItemsCount::from(1_u32))
+        .with_inputs(|| create_random_matrix::<2, 2, _>(&mut rng))
+        .bench_local_values(|a| black_box(a.inverse()));
+}
+
+#[cfg(not(target_arch = "wasm32"))]
```
joaander:
```suggestion
```

hoomd-linear-algebra/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+//! This crate places an emphasis on generality and simplicity, with optimization efforts
+//! targeted at small matrixes. Some complex routines (SVD, matrix inversion, etc.) will
+//! only be implemented for certain shapes, and generally consist of specialized algorithms
+//! optimal for those inputs.
+//!
+//! This crate should not be considered a replacement for a dedicated linear algebra library
+//! like [faer-rs](https://github.com/sarah-quinones/faer-rs.git) or
+//! [nalgebra](https://github.com/dimforge/nalgebra). Instead, it can be used as a simple
+//! and lightweight dependency for matrix methods common to molecular simulation and
+//! analysis.
```
joaander:
```suggestion
//! like [faer-rs] or [nalgebra]. Instead, it can be used as a simple
//! and lightweight dependency for matrix methods common to molecular simulation and
//! analysis.
//!
//! [faer-rs]: https://github.com/sarah-quinones/faer-rs.git
//! [nalgebra]: https://github.com/dimforge/nalgebra
```
This makes it easier to read the comment directly and reword things later.

hoomd-linear-algebra/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    /// Multiply a matrix by a general matrix RHS.
+    #[must_use]
+    fn matmul(&self, rhs: &RHS) -> Self::Output;
+}
+
+/// General implementation for size and container-agnostic matrixes.
+///
+/// This trait is designed to function with row-major ordering, but this is not strictly
+/// required for correct functionality.
```
joaander: Expand this documentation to list (in words) all the operations that can be applied to a general matrix. Example code blocks for each would be nice.

You say "designed to function with row-major ordering". Does that mean `Index` for one `GeneralMatrix` might be row-major while it might be column-major for another?

hoomd-linear-algebra/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    + Index<(usize, usize), Output = f64>
+    + Neg
+{
+    /// Fill a matrix with zeros.
+    #[must_use]
+    fn zeros() -> Self;
+
+    /// Return a matrix where every element is equal to val.
+    #[must_use]
+    fn full(val: f64) -> Self;
```
joaander:
```suggestion
    /// Return a matrix where every element is equal to value.
    #[must_use]
    fn full(value: f64) -> Self;
```
I read this as "Val" (as in the the name of a person).

hoomd-linear-algebra/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+}
```
joaander, line 325: You can index a general matrix, but you can't get its dimensions?

hoomd-linear-algebra/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    #[must_use]
+    fn matmul(&self, rhs: &RHS) -> Self::Output;
+}
+
+/// General implementation for size and container-agnostic matrixes.
+///
+/// This trait is designed to function with row-major ordering, but this is not strictly
+/// required for correct functionality.
+pub trait GeneralMatrix:
+    Sized
```
joaander: Is the `Sized` bound required? I thought that the compiler automatically added this to all sized types.

hoomd-vector/src/cartesian.rs
```diff
@@ -698,6 +708,52 @@ impl<const N: usize> Rotate<Cartesian<N>> for RotationMatrix<N> {
...
+                    result[(n, k)] += self.rows()[n][m] * rhs[(m, k)];
+                }
+            }
+        }
+
+        result
+    }
+}
+
+impl<const N: usize> Diagonal for Cartesian<N> {}
```
joaander: This needs some documentation explaining what it means and how it could be used.

If I read the code correctly, this allows a `Cartesian<N>` to be used as a diagonal matrix. In math notation, normally we think of vectors as row or column vectors. I would omit this implementation and require callers to explicitly state that they are forming a diagonal matrix from the elements of a vector if that is what they want to do:
```
from_diag(vector.coordinates);
```

hoomd-linear-algebra/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+/// let mat = Matrix22::full(5.0);
+/// assert_eq!(mat.matmul(&Matrix22::identity()), mat);
+///
+/// let diag = Matrix22::from_diag(&[3.0, 2.0]);
+///
+/// assert_eq!(
+///     mat.matmul(&diag),
+///     mat.matmul(&DiagonalMatrix { rows: [3.0, 2.0] })
+/// );
+/// ```
```
joaander: These examples would be more clear if we wrote out the 2nd argument of `assert_eq!` explicitly: e.g. `[[5.0, 5.0], [5.0, 5.0]].into()`.

hoomd-linear-algebra/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    #[must_use]
+    fn zeros() -> Self;
+
+    /// Return a matrix where every element is equal to val.
+    #[must_use]
+    fn full(val: f64) -> Self;
+}
+
+/// Marker trait to indicate a sequence of values can be read as a diagonal matrix.
+pub trait Diagonal: Index<usize, Output = f64> {}
```
joaander: I see that there is a type `DiagonalMatrix` that has symmetry with `GeneralMatrix` and `SquareMatrix` names.

I don't understand what purpose the marker `Diagonal` serves. It is implemented for:
* `DiagonalMatrix<N>`
* `[f64; N]`
* `Cartesian<N>`
Are there other indexable types that we need to prevent users from passing to `from_diag` or `compute_quadratic_form` for safety?

hoomd-vector/src/cartesian.rs
```diff
@@ -698,6 +708,52 @@ impl<const N: usize> Rotate<Cartesian<N>> for RotationMatrix<N> {
...
+
+#[allow(dead_code, reason = "TODO: Used in other branch.")]
```
joaander: It's find that this code is used on another branch. But how useful is it as a private fn? If you mark these `pub`, then the dead code warning will go away under the assumption that other code will use them.

These methods also need unit tests.

hoomd-linear-algebra/src/matrix.rs
```diff
@@ -0,0 +1,1309 @@
+// Copyright (c) 2024-2025 The Regents of the University of Michigan.
+// Part of hoomd-rs, released under the BSD 3-Clause License.
+
+use std::ops::{Add, Index, IndexMut, Mul, Neg, Sub};
+
+use crate::{Diagonal, GeneralMatrix, Invertible, MatMul, QuadraticForm, SquareMatrix};
+// use hoomd_vector::{Cartesian, RotationMatrix};
```
joaander:
```suggestion
```

hoomd-linear-algebra/src/matrix.rs
```diff
@@ -0,0 +1,1309 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub struct Matrix<const N: usize, const M: usize> {
+    /// The elements of the matrix
+    pub rows: [[f64; M]; N],
+}
+/// A square, diagonal matrix with N rows and N columns.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct DiagonalMatrix<const N: usize> {
+    /// The elements of the diagonal of the matrix
+    pub rows: [f64; N],
```
joaander:
```suggestion
    pub diagonals: [f64; N],
```
This doesn't make sense to call `rows`.

hoomd-linear-algebra/src/matrix.rs
```diff
@@ -0,0 +1,1309 @@
...
+    fn index(&self, index: (usize, usize)) -> &f64 {
+        let (i, j) = index;
+        if i == j { &self.rows[i] } else { &0.0 }
+    }
+}
+/// Index the rows and columns of a [`Matrix`]
+///
+/// Indices for [`Matrix`] types are zero-indexed and reflect the indexing pattern of
+/// the underlying data. This results in the pattern `(row, column)`, which mirrors the
+/// behavior of Numpy and similar array languages.
```
joaander: I would make this a stronger statement and document up in the `GeneralMatrix` trait as well. `Index` should *always* index by `(row, column)` regardless of the underlying storage.

hoomd-linear-algebra/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+
+/// General implementation for size and container-agnostic matrixes.
+///
+/// This trait is designed to function with row-major ordering, but this is not strictly
+/// required for correct functionality.
+pub trait GeneralMatrix:
+    Sized
+    + Mul<f64, Output = Self>
+    + Add<Self, Output = Self>
+    + Index<(usize, usize), Output = f64>
```
joaander, line 291: Why no `IndexMut`?

hoomd-linear-algebra/src/matrix.rs
```diff
@@ -0,0 +1,1309 @@
...
+        Self {
+            rows: std::array::from_fn(|_| 0.0),
+        }
+    }
+    #[inline]
+    fn full(val: f64) -> Self {
+        Self {
+            rows: std::array::from_fn(|_| val),
+        }
+    }
```
joaander, line 78: The off diagonal elements are 0, so `full` does not fulfill its promise and create a matrix where every element is equal to the given value.

To implement `GeneralMatrix` for `DiagonalMatrix`, you need to break `full` out into a separate `Full` trait.

hoomd-linear-algebra/src/matrix.rs
```diff
@@ -0,0 +1,1309 @@
...
+    #[inline]
+    fn identity() -> Self {
+        Self {
+            rows: std::array::from_fn(|i| std::array::from_fn(|j| if i == j { 1.0 } else { 0.0 })),
+        }
+    }
+}
+
+impl<const N: usize> DiagonalMatrix<N> {
+    /// Return a dense view of the diagonal matrix, with zeros on the off-diagonals.
```
joaander:
```suggestion
    /// Return a dense copy of the diagonal matrix, with zeros on the off-diagonals.
```

hoomd-linear-algebra/src/matrix.rs
```diff
@@ -0,0 +1,1309 @@
...
+//             for k in 0..K {
+//                 for m in 0..M {
+//                     result.rows[n][k] += self.rows[n][m] * rhs.rows[m][k];
+//                 }
+//             }
+//         }
+
+//         result
+//     }
+// }
```
joaander:
```suggestion
```
Implemented above.

hoomd-linear-algebra/src/matrix.rs
```diff
@@ -0,0 +1,1309 @@
...
+                row[j] = self.rows[i][j] * rhs[j];
+            }
+        }
+        result
+    }
+}
+
+impl<const N: usize, const M: usize> Matrix<N, M> {
+    /// Interchange the rows and columns of matrix `A` such that `A.transpose()[(j, i)] = A[(i, j)]`
+    #[allow(dead_code, reason = "No use case yet.")]
```
joaander: It is used in `map_col` below.

Transpose is useful to the caller. Mark it `pub`.

hoomd-linear-algebra/src/matrix.rs
```diff
@@ -0,0 +1,1309 @@
...
+//             f,
+//             "[{}]",
+//             self.rows
+//                 .map(|row| format!("[{}]", row.map(f64::to_string)))
+//                 .into_iter()
+//                 .collect::<Vec<String>>()
+//                 .join("\n ")
+//         )
+//     }
+// }
```
joaander:
```suggestion
```
Reimplement this or remove it.

hoomd-linear-algebra/src/matrix.rs
```diff
@@ -0,0 +1,1309 @@
...
+
+impl Matrix<2, 2> {
+    /// Decompose a [`Matrix22`] into a rotation `U`, a scaling`Σ`, and a second rotation`Vt` such that `A=UΣVt`.
+    ///
+    /// This implementation is based on the math in 10.1109/38.486688, and ensures good
+    /// (but not optimal) numerical stability. For certain pathological inputs,
+    /// preconditioning the matrix could provide a benefit in numerical stability.
+    ///
+    /// We define all singular values to be positive.
```
joaander:
```suggestion
    /// Decompose a [`Matrix22`] into a rotation `U`, a scaling`Σ`, and a second rotation`Vt` such that `A=UΣVt`.
    ///
    /// This implementation is based on the math in 10.1109/38.486688, and ensures good
    /// (but not optimal) numerical stability. For certain pathological inputs,
    /// preconditioning the matrix could provide a benefit in numerical stability.
    ///
    /// `svd` sets all singular values positive.
```
Simplify the first line into a sentence, move the math into the longer description and use LaTeX formatting.

hoomd-linear-algebra/src/matrix.rs
```diff
@@ -0,0 +1,1309 @@
...
+//             }
+//         }
+
+//         result
+//     }
+// }
+
+impl<const N: usize, const M: usize> MatMul<DiagonalMatrix<M>> for Matrix<N, M> {
+    type Output = Matrix<M, M>;
+    /// Scale each column of a [`Matrix`] by the corresponding element in a [`DiagonalMatrix`].
```
joaander: This is correct, but worded this way it makes it sound like this is not a matrix multiply. I would keep the same general first line description even for specialized implementations of a method.

hoomd-linear-algebra/src/matrix.rs
```diff
@@ -0,0 +1,1309 @@
...
+        case::identity(Matrix22::identity().rows),
+        case::mixed_sign([[1.0, -2.0], [3.0, 4.0]]),
+        case::det_zero([[12.0, 2.0], [4.0, 0.0]]),
+        case::large_range([[1000.0, 0.0], [0.0, 1e-4]]),
+        case::jordan_block([[1.0, 1.0], [0.0, 1.0]]),
+        case::full_ones(Matrix22::full(1.0).rows),
+        case::shear([[1.0, 2.0], [0.0, 1.0]]),
+        case::nilpotent([[0.0, 1.0], [0.0, 0.0]]),
+        case::scaling([[2.0, 0.0], [0.0, 3.0]]),
+        /* None of these examples work using the fast algorithm.*/
```
joaander: These don't work because they are degenerate? Or because of a problem with the fast algorithm?

hoomd-linear-algebra/src/matrix.rs
```diff
@@ -0,0 +1,1309 @@
...
+
+        (u, singular_values, vt)
+    }
+}
+
+impl Matrix<4, 4> {
+    /// Compute the determinant of a 4x4 symmetric matrix
+    #[expect(clippy::many_single_char_names, reason = "clarity")]
+    #[allow(dead_code, reason = "TODO: will be used elsewhere.")]
+    pub(crate) fn det44_symmetric(&self) -> f64 {
```
joaander: We would want to implement this on a specialized symmetric matrix type for safety.

joaander (changes requested):
Thanks! This will be very useful.

I've made a number of suggestions to improve the code. I've also asked a lot of questions where I don't understand the code or documentation.

The more recent functions (fold, map, ...) have complete documentation with examples. We should apply the same treatment to all the types, traits, and methods.

joaander:
TODO:
* [x] Add a link to `api.md` to the new crate's documentation.

hoomd-linear-algebra/src/matrix.rs
janbridley: The fast algorithm computes a correct solution (reconstructs matrix properly, within some precision), but does not recover the 'canonical' solution one would expect. I've updated the comment for this, and we test those cases against another implementation of the same algorithm

hoomd-linear-algebra/benches/linalg.rs
joaander: Please add a comment explaining why this benchmark tests a function implemented only in the benchmark file. The same applies to det44 below.

hoomd-linear-algebra/src/diagonal.rs
```diff
@@ -0,0 +1,257 @@
...
+
+use std::ops::{Add, Index, IndexMut, Mul, Neg, Sub};
+
+use crate::{Diagonal, GeneralMatrix, matrix::Matrix};
+
+/// A square, diagonal matrix with N rows and N columns.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct DiagonalMatrix<const N: usize> {
+    /// The elements of the diagonal of the matrix
+    pub rows: [f64; N],
```
joaander: I still think that `rows` is not the correct name. This is a public field and therefore part of the API. Use `elements` or `diagonal_elements`, unless you have a better suggestion.

hoomd-linear-algebra/src/diagonal.rs
```diff
@@ -0,0 +1,257 @@
...
+
+impl<const N: usize> GeneralMatrix for DiagonalMatrix<N> {
+    #[inline]
+    fn zeros() -> Self {
+        Self {
+            rows: std::array::from_fn(|_| 0.0),
+        }
+    }
+    #[inline]
+    fn full(val: f64) -> Self {
```
joaander:
```
let mat = DiagonalMatrix::full(3.0);
assert_eq!(mat[(0, 2)], 3.0); # Fail!
```
The assert should be true based on the documentation of `full`:
> Return a matrix where every element is equal to val.

It is *not possible* to implement `full` for `DiagonalMatrix`.

hoomd-linear-algebra/src/diagonal.rs
```diff
@@ -0,0 +1,257 @@
...
+            rows: std::array::from_fn(|_| val),
+        }
+    }
+}
+
+impl<const N: usize> DiagonalMatrix<N> {
+    /// Return a dense view of the diagonal matrix, with zeros on the off-diagonals.
+    #[must_use]
+    #[inline]
+    pub fn as_dense(&self) -> Matrix<N, N> {
```
joaander: This should be `to_dense` and should go owned -> owned. See the Rust API guidelines for details: https://rust-lang.github.io/api-guidelines/naming.html?highlight=as_#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

hoomd-linear-algebra/src/diagonal.rs
```diff
@@ -0,0 +1,257 @@
...
+        Matrix::<N, N>::from_diag(&self.rows)
+    }
+}
+
+/// Compute the elementwise scalar multiplication of a [`DiagonalMatrix`]
```
joaander: This is what it does, yes. But the way this is worded makes it sound like it is *not* a matrix-matrix multiplication (which it is).

I would delete this misleading doc comment. The `mul` method below will get a copy of the trait's documentation.
```suggestion
```

hoomd-linear-algebra/src/diagonal.rs
```diff
@@ -0,0 +1,257 @@
...
+    type Output = Self;
+
+    #[inline]
+    fn mul(self, rhs: f64) -> Self {
+        Self {
+            rows: self.rows.map(|r| r * rhs),
+        }
+    }
+}
+/// Compute the elementwise negation of a [`DiagonalMatrix`]
```
joaander:
```suggestion
```
I don't think we need to document `impl` blocks in general unless there is something special about the contents or the way generics are handled inside.

hoomd-linear-algebra/src/lib.rs
joaander: Not yet resolved.

hoomd-linear-algebra/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+///
+/// # Example
+/// ```
+/// use hoomd_linear_algebra::{
+///     GeneralMatrix, MatMul, SquareMatrix,
+///     matrix::{DiagonalMatrix, Matrix22},
+/// };
+///
+/// let mat = Matrix22::full(5.0);
+/// assert_eq!(mat.matmul(&Matrix22::identity()), mat);
```
joaander:
```suggestion
/// let a = Matrix22::full(5.0);
/// assert_eq!(a.matmul(&Matrix22::identity()), &[[5.0, 5.0], [5.0, 5.0]].into());
```
A mat is an object that you place on the floor and stand on, usually rectangular and made of fiber or rubber.

I think generic names like "a" are sufficient for documentation examples.

hoomd-linear-algebra/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+/// );
+/// ```
+pub trait MatMul<RHS>
+where
+    RHS: GeneralMatrix,
+{
+    /// The type of the output matrix.
+    ///
+    /// This type is likely to be Self for dynamically sized [`GeneralMatrix`] types, but
+    /// will necessarily be different for statically allocated rectangular matrixes.
```
joaander:
```suggestion
```
I don't think this information is accurate. There are no dynamically sized matrices in this library.

hoomd-linear-algebra/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+/// let diag = Matrix22::from_diag(&[3.0, 2.0]);
+///
+/// assert_eq!(
+///     mat.matmul(&diag),
+///     mat.matmul(&DiagonalMatrix { rows: [3.0, 2.0] })
+/// );
+/// ```
+pub trait MatMul<RHS>
+where
+    RHS: GeneralMatrix,
```
joaander:
```suggestion
```
This unnecessarily constrains implementations of this trait. There may be types that are not `GeneralMatrix` that still have valid matrix multiplication semantics.

hoomd-linear-algebra/src/lib.rs
joaander: This comment has not been addressed.

hoomd-linear-algebra/src/lib.rs
joaander: This comment has not been addressed.

hoomd-linear-algebra/src/lib.rs
joaander: This comment has not been addressed?

hoomd-linear-algebra/src/lib.rs
joaander: This comment has not been addressed.

hoomd-linear-algebra/src/lib.rs
joaander: This comment has not been addressed.

hoomd-linear-algebra/src/lib.rs
joaander: This comment has not been addressed. I still don't understand why the caller needs some sort of special trait to treat some objects as diagonal matrices. They can either a) form them explicitly or b) use an ad-hoc conversion like `DiagonalMatrix::from_array`.

hoomd-vector/src/cartesian.rs
```diff
@@ -698,6 +708,52 @@ impl<const N: usize> Rotate<Cartesian<N>> for RotationMatrix<N> {
...
+    #[inline]
+    fn matmul(&self, rhs: &Matrix<N, K>) -> Self::Output {
+        let mut result = Self::Output::zeros();
+        for n in 0..N {
+            for k in 0..K {
+                for m in 0..N {
+                    result[(n, k)] += self.rows()[n][m] * rhs[(m, k)];
+                }
+            }
+        }
```
joaander: Reduce code duplication by using `From` to convert and then invoking the existing `MatMul` for Matrix-Matrix multiplication.

hoomd-vector/src/cartesian.rs
joaander: This comment has not been addressed.

hoomd-vector/src/cartesian.rs
joaander: This comment has not been addressed.

hoomd-linear-algebra/src/matrix/ops.rs
```diff
@@ -0,0 +1,351 @@
...
+/// ```
+/// use hoomd_linear_algebra::matrix::Matrix;
+/// let rows = [[1.0, 2.0], [3.0, 4.0], [5.0, 6.0]];
+/// let mat = Matrix { rows };
+/// assert_eq!(mat[(0, 1)], rows[0][1]);
+/// assert_eq!(mat[(2, 1)], 6.0);
+/// assert_eq!(mat[(1, 1)], 4.0);
+/// // Out-of-bounds: would panic!
+/// // mat[(3, 0)];
+/// ```
```
joaander: I find that documentation on the `impl` block appears in a weird place in the documentation. I will move these blocks to the method that performs the documented operation.

hoomd-linear-algebra/src/matrix.rs
joaander: This comment has not been addressed.

joaander (commented):
I'll go ahead and implement the obvious fixes for the remaining unresolved comments. I'll let you know what (if any) further questions remain at that point.

hoomd-linear-algebra/src/lib.rs
joaander: Odd. Remove the Sized and it won't compile. We use exactly the same syntax for `Vector` and it works without Sized.... I don't understand. We'll leave it in.

hoomd-linear-algebra/src/lib.rs
joaander: In general, we should discuss what the complete set of matrix operations should be.

janbridley: Agreed on the latter - I left this open so that I could implement n_rows and n_cols as const functions where possible, but maybe we make those (or similar names) part of GeneralMatrix and leave a Numpy-style `shape` that could be a const fn

joaander: We can't require `IndexMut` because that cannot be implemented for `DiagonalMatrix`.

hoomd-linear-algebra/src/matrix.rs
```diff
@@ -0,0 +1,1386 @@
...
+                .collect::<Vec<_>>()
+                .join(",\n ")
+        ))
+    }
+}
+impl Matrix<2, 2> {
+    /// Decompose a [`Matrix22`] into a rotation, a scaling, and a second rotation.
+    ///
+    /// ```math
+    /// A = U Σ V^\top;
```
joaander: Replace this with LaTeX math.

hoomd-linear-algebra/src/matrix.rs
```diff
@@ -0,0 +1,1386 @@
...
+
+        (u, singular_values, vt)
+    }
+}
+
+impl Matrix<3, 3> {
+    /// Decompose a [`Matrix33`] into a rotation, a scaling, and a second rotation.
+    ///
+    /// ```math
+    /// A = U Σ V^\top;
```
joaander: Replace this with LaTeX math.

hoomd-linear-algebra/src/matrix.rs
```diff
@@ -0,0 +1,1386 @@
...
+    /// ```
+    /// This implementation is based on the method described by `McAdams` et al. in
```
joaander: Cite with a link.

joaander (commented):
OK, I've implemented my suggestions directly.

TODO:
* [x] There are a couple math equations remaining to fix. 
* [ ] We should discuss if the `GeneralMatrix` supertraits are complete or not.
* [x] Explain why the benchmarks benchmark their own methods.
* [x] Merge trunk into this branch.

hoomd-linear-algebra/src/lib.rs
joaander: I'm not sure that `shape` could be a const fn. You implemented these traits so that they *could* potentially be used for heap allocated matrices if the need arises.

Should we just leave `shape` out of `GeneralMatrix` for now then? Or implement it non-const and trust that the compiler to inline it away? `const` is only needed if we want to ensure that it can be used in a `const` context - which we can for `Matrix` by using `N` and `M` directly.

joaander (approved):
Thanks! I made a final pass on the docs and updated some names for consistency / symmetry.

The last open conversation regards `shape`. Should we call it `dimensions` as that is the term more often used in the mathematical literature? Should we move it to `GeneralMatrix` or not? I can go either way on these, so I leave it up to you.

joaander (approved):
Thanks!

PR #70, opened 2025-09-11
Create macrostate module

josephburkhart:
Adds a new module that can be used to track global system parameters. Suggestions are welcome regarding trait names and module location. This addition will enable refactoring of method signatures in the `md` and `mc` crates so that associated types will no longer be necessary.


josephburkhart:
I just fixed all of the pre-commit issues issues except for the snake-case one. I personally think that we should have the temperature field be named `kT`, rather than `kt` or `k_t`, since that more clearly distinguishes it from time-related quantities like `dt`. I don't think there has been a group decision on this, though.

PR #76, opened 2025-09-17
Format documentation examples.

joaander:
Unfortunately, `rustfmt` doesn't format block doc comments unless you prefix every line with a `*` like this:

<pre>
    /** Construct an empty microstate with open boundary conditions.
     *
     * The microstate starts at step 0, substep 0, random number seed 0,
     * and has no bodies.
     *
     * # Example
     *
     * ```
     * use hoomd_microstate::Microstate;
     * # use hoomd_microstate::{Body, property::Point};
     * # use hoomd_vector::Cartesian;
     *
     * let mut microstate = Microstate::new();
     * assert_eq!(microstate.step(), 0);
     * assert_eq!(microstate.substep(), 0);
     * assert_eq!(microstate.seed(), 0);
     * assert_eq!(microstate.bodies().len(), 0);
     * assert_eq!(microstate.sites().len(),0);
     * # microstate.add_body(Body::point(Cartesian::from([0.0, 0.0])));
     * ```
     */
</pre>

There seems to be a lot of inconsistency in the rust tooling surrounding this type of comment. The `*` are not supposed to be there (https://github.com/rust-lang/rust/issues/92872), yet `rustfmt` will not format the code unless they are. `rustdoc` at times has stripped the leading `*` and included them as part of the markdown at other times. Also, the latest version of `helix` doesn't recognize `/**` styles comments properly for syntax highlighting.

The best solution I can come up with is to enable `normalize_comments` which rewrites all the doc comments with `///` and then formats them.

### DIFF
--- a/hoomd-bevy/src/lib.rs
+++ b/hoomd-bevy/src/lib.rs
@@ -103,14 +103,14 @@ const CAMERA_ZOOM_SPEED: f32 = 50.0;
 /// A menu to control common settings (steps per second limit, camera speed, etc.)
 ///
 /// The caller must:
-/// Provide type that implements [`Simulation`].
-/// Add a `sync` `Update` system that populates (and removes) entities for
-/// rendering. See [`representation`] for helper code.
+/// * Provide type that implements [`Simulation`].
+/// * Add a `sync` `Update` system that populates (and removes) entities for
+///   rendering. See [`representation`] for helper code.
 ///
 /// The caller may optionally:
-/// Add UI to the upper right corner of the screen.
-/// Implement custom controls.
-/// Add lines to the [`HelpText`] entity.
+/// * Add UI to the upper right corner of the screen.
+/// * Implement custom controls.
+/// * Add lines to the [`HelpText`] entity.
 ///
 /// To keep individual example scripts short and understandable, `hoomd-bevy` should
 /// implement as much common code as possible.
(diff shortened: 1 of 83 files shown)


hoomd-bevy/src/lib.rs
```diff
@@ -152,15 +150,14 @@ pub enum MenuState {
...
-    * Left click and drag to pan.
-    * Scroll to zoom.
-    */
+    /// Two dimensional top down camera showing the xy plane.
+    ///
+    /// The single field sets the height of the visible area. The width is set
+    /// automatically based on the window dimensions.
+    ///
+    /// Controls:
+    /// Left click and drag to pan.
```
joaander: Unfortunately, we need to review every rewritten comment because `normalize_comments` eats leading `*` characters.

janbridley:
As long as it's all processed automatically this seems like a reasonable approach. We can always update the config once they've stabilized the proper syntax. 

I ran into some weirdness recently where certain doc comment indentations would be compiled as doc examples even if they weren't in an example or code block, so hopefully this might address that as well

janbridley (approved):

joaander:
> I ran into some weirdness recently where certain doc comment indentations would be compiled as doc examples even if they weren't in an example or code block, so hopefully this might address that as well

Yes I think it will. The fundamental problem with `/**` style comments is that there is no one rule to strip leading whitespace (or other characters like * which are commonly used) that will match every developers' intent. Leave the wrong number of whitespace characters on a line and then markdown will interpret it as a code block when rustdoc then interprets as an example to compile.

With `///` there is a clear starting point for each line of documentation and this is not an issue.

joaander:
I ran into the same whitespace removal problem in HOOMD-blue. Python 3.13 changed the way that leading whitespace was removed from docstrings, which broke Sphinx because reStructureText is much more indentation-sensitive than markdown.

PR #87, opened 2025-09-22
Add dimensionality getter to Vector

josephburkhart:
This PR adds a new method to the `Vector` trait definition. The new method returns a `usize` that specifies the dimensionality of the vector-space. For the type `Cartesian<N>`, the method simply returns `N`. This functionality is useful e.g. in the `md` crate, where integration methods require knowledge of the number of degrees of freedom in a system.


joaander (approved):
Thanks!

PR #92, opened 2025-09-23
Add caching to CI scripts

janbridley:
This definitely improves performance and does not seem to result in cache collisions between different versions of rust, although I'd love an extra pair of eyes to validate that. I used `${{ runner.name }}` for the test CI to ensure a separate cache would be kept for each matrix entry. This causes cache misses on jet stream (as the cpu index is included as part of the name it seems?) but should be fine on GitHub actions, where the naming is more consistent. It probably doesn't matter if we use `${{ runner.name }}` or `${{ runner.os }}` but I figured I'd err on the side of caution.

### DIFF
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -22,6 +22,14 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+    - name: Cache
+      uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+          target
+        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
     - name: Configure Rust
       uses: dtolnay/rust-toolchain@6d653acede28d24f02e3cd41383119e8b1b35921
       with:
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -50,6 +50,14 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+    - name: Cache
+      uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+          target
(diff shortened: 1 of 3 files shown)


janbridley:
We could go faster by creating a matrix over the `cargo check|clippy` args, but that multiplies the number of style CI runners by 4. Caching seems to cut the runtime to under a minute so I'm reasonably happy - @joaander let me know your thoughts!

joaander:
I don't think `runner.name` makes sense to use in the hash key. What if the job running on `actions-runner-cpu1` previously was a `dev` build and this one is `release`? It will need to rebuild everything. The hash key should be based on the matrix so that you pull in the same data used in the previous run:
```
key: ${{ github.workflow }}-${{ matrix.profile }}-rust-${{ matrix.rust }}-${{ matrix.os }}-${{ matrix.target }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
It is a good idea to include the workflow name so that caches don't mix between the style check and test workflows, although the style check workflow will likely have a different set of keys in the name.

I previously had caching enabled on hoomd-rs, but needed to disable it. GitHub gives us only 10 GB of cache (though we can pay $$ for more). hoomd-rs builds use about ~1.9 GB per matrix element (https://github.com/glotzerlab/hoomd-rs/actions/caches) so we overflow the cache with one push. When we open-source hoomd-rs, we will add mac and windows matrix elements and further increase the size of the cache needed.

I can't tell because they have already been evicted, but it is possible that the style check workflow produces a much smaller cache than the test workflow.

Options to improve CI total time to completion within the cache limits:
* Run style checks in a matrix.
* Run style checks on jetstream2.
* Investigate the size of the various cached directories:
  * How big is `.cargo/registry`. Do we really need it to reduce the time or do we just let cargo download it from `crates.io` each time?
  * We need `target` to reduce the build times, but how much of `target` do we need?

I'll look into `target` and see what we can do.

joaander:
There are long issue threads dating back to 2018 where people have been asking for cargo features to make CI caching easier. There is no builtin support.

This action wraps `actions/cache` and cleans unneeded files from: https://github.com/Swatinem/rust-cache. I'm trying it now.

joaander:
Using [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache):

* check-clippy: 5 minutes [miss] / 45s [hit]
* unit test native: 3 minutes [miss] / 1 minute [hit]
* unit test wasm32: 2 minutes [miss] / 25 seconds [hit]

The caches are 300-500 MB per matrix element which is gives us headroom to add more matrix elements when needed.

joaander (approved):
Thanks!

When we migrate the examples to GitHub pages, we can use the cache to greatly speed up the interactive example builds. There is nothing we can do right now to make readthedocs faster.

PR #102, opened 2025-10-05
Lennard jones gauss

mthran:
Added a length scale to the lennard-jones-gauss potential.


joaander (approved):
Thanks!

I see that this also fixes the bug in `force` that @shihkual mentioned.

PR #130, opened 2025-11-12
Add collision tests for elongated bodies

janbridley:
Infinite cylinder intersection test, useful as a fast early exit for HPMC with elongated shapes. This requires only a few operations, and I expect it is faster when the intersection volume of the cylinder + neighbor list ball is significantly less than that of the bounding sphere of a shape - this check takes ~4x as long as the sphere, so that seems a reasonable guess. I did not implement the finite cylinder check, as honestly it doesn't seem too interesting -- we can revisit in the future if desired.

This implementation is based on [Collision detection of cylindrical rigid bodies for motion planning](https://doi.org/10.1109/ROBOT.2006.1641925).
```
├─ infinite_cylinder         1.854 ns      │ 2.668 ns      │ 2.098 ns      │ 2.095 ns      │ 100     │ 204800
│                            539.3 Mitem/s │ 374.8 Mitem/s │ 476.6 Mitem/s │ 477.3 Mitem/s │         │
╰─ sphere_fast_nd                          │               │               │                               4.219 Gitem/s │ 827.8 Mitem/s │ 3.968 Gitem/s │ 3.676 Gitem/s │         │
   ├─ 3                      0.532 ns      │ 0.593 ns      │ 0.562 ns      │ 0.558 ns      │ 100     │ 409600
   │                         1.879 Gitem/s │ 1.686 Gitem/s │ 1.779 Gitem/s │ 1.792 Gitem/s │         │
```

This also includes code for N-Capsules, based on closest point on lines. This is in preparation for more complex swept sphere volumes.
```
├─ capsule_fast_3d           14.91 ns      │ 26.96 ns      │ 15.16 ns      │ 15.27 ns      │ 100     │ 51200
│                            67.02 Mitem/s │ 37.08 Mitem/s │ 65.95 Mitem/s │ 65.44 Mitem/s │         │
├─ capsule_xenocollide_3d    92.8 ns       │ 393.5 ns      │ 126.9 ns      │ 128.1 ns      │ 100     │ 6400
│                            10.77 Mitem/s │ 2.54 Mitem/s  │ 7.875 Mitem/s │ 7.802 Mitem/s │         │

```

Also note it looks like my 2d ellipse code is horribly inefficient -  I made the easy fixes I could, but we need to revisit in a later PR.

PR #131, opened 2025-11-12
Add random small quaternion displacement

janbridley:
## Random small quaternions

Useful for trial moves, implemented based [Karney 2007](doi.org/10.1016/j.jmgm.2006.04.002) who is cited by Frenkel and many others. The distribution behaves how I expect it would, although I don't have a good intuition for how to precisely quantify the outputs. I found a good heuristic test that passes for a large variety of inputs. Note that values of a > pi/2 or so will fail the test as sample mean is no longer guaranteed to be at the origin.

quat_distribution

**Possible TODO**: we generate a quaternion that is normalized, then re-normalize the output because quaternion->versor conversions are protected. Ideally we do not have to renormalize, as the norm should always be 0. Note that not normalizing before converting to versor leads to a different distribution than in Karney.

One fix is enabling #[repr(transparent)] on Versor and then std::mem::transmute


hoomd-mc/src/rotate/versor.rs
```diff
@@ -0,0 +1,172 @@
...
+            // Lift the normally distributed values to SO(3) with the exponential map
+            let half_theta = 0.5 * theta;
+            let w = half_theta.cos();
+
+            // If theta is near 0, do not compute the value. TODO: detailed balance?
+            let v_factor = if theta < _SAFE_THETA {
+                0.5
+            } else {
+                half_theta.sin() / theta
+            };
```
joaander:
```suggestion
            let v_factor = if theta == 0.0 {
                0.5
            } else {
                half_theta.sin() / theta
            };
```
I think to have detailed balance here, you would need to reject any sample with theta < _SAFE_THETA. That would treat the positive and negative rotation cases symmetrically.

I also don't see any problems removing `_SAFE_THETA` and using an `== 0.0` check. No matter what the exponent is, the mantissa should provide sufficient relative accuracy to generate a valid quaternion, except in the case that it is identically 0. You could reject those, or generate an identity versor as you do here.

```python
>>> theta = 1e-300
>>> math.sin(theta/2) / theta
0.5
```

hoomd-mc/src/rotate/versor.rs
```diff
@@ -0,0 +1,172 @@
...
+
+            let v = s * v_factor;
+
+            // We are normalized by construction, so call the tuple initializer
+            return (Quaternion::from([w, v[0], v[1], v[2]])).to_versor_unchecked(); // TODO: no need to re-normalize!
```
joaander, line 69: If you want to avoid the repeat normalization, move `VersorDisplacement` to the `hoomd-vector` crate and make it a public API. Add a `pub(crate)` API to `Versor` that allows constructing one from an already normalized quaternion:
```rust
pub(crate) fn from_normalized_quaternion
```

hoomd-vector/src/quaternion.rs
```diff
@@ -571,6 +576,32 @@ impl Versor {
...
+    ///
+    /// $`d : \mathbb{H} \times \mathbb{H} \to \mathbb{R}^+, \quad d(q_0, q_1) = 1 - |q_0 \cdot q_1 |`$
+    ///
+    /// This has less geometric meaning than the [`arc_distance`](Versor::arc_distance) metric. However, it
+    /// is much faster while still obeying the triangle inequality and the axiom
+    /// $`d(q_0, q_1) = d(q_1, q_0)`$. This metric always lies in the range
+    /// $`[0, 1]`$, and is symmetric such that $`d(q, q)`$ = $`d(q, -q)`$.
+    #[inline]
+    #[must_use]
+    pub fn half_euclidean_norm_squared(&self, other: &Self) -> f64 {
```
joaander, line 602: This appears to be unused except for a benchmark. Do you have plans to use it on other branches?

joaander (commented):
Thanks! I have a few suggestions to improve this code. Let me know what you think.

hoomd-mc/src/rotate/versor.rs
joaander: I'm no longer sure that `== 0.0` is the correct check. There might be some corner case where a non-zero theta results in an overflowed `v_factor`. It may be more robust to explicitly check and handle the overflow case (as I suggested in https://github.com/glotzerlab/hoomd-rs/pull/130#discussion_r2550448783):
```suggestion
            let mut v_factor = half_theta.sin() / theta;
            if !v_factor.is_finite() {
                v_factor = 0.5;
            }
```

hoomd-vector/src/quaternion.rs
janbridley: I'm planning on using this for order parameter calculations down the line -- are we ok leaving the dead code for a bit?

joaander: Yes, that is fine.

PR #181, opened 2026-01-28
Update vector distribution docs

### LINKED ISSUE #117: Document vector random distributions better.
The distributions module should mention `StandardUniform`. The overview docs should also mention `Ball`.

janbridley:
@joaander let me know if this sufficiently addresses what you were hoping for -- I'm happy to add more!

Fixes #117

### DIFF
--- a/hoomd-vector/src/lib.rs
+++ b/hoomd-vector/src/lib.rs
@@ -178,8 +178,8 @@
 //! ```
 //!
 //! The [`Ball`](crate::distribution::Ball) distribution samples vectors from
-//! the surface and interior of an `n-Ball`, the set of all points whose distance from
-//! the origin is in `[0, 1)`.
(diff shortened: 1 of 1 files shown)


joaander (approved):
Thanks! This is exactly what I was looking for.

On review, I noticed that the links to `rand` are broken. It seems that `rand` is no longer hosting their docs on GitHub pages. Due to https://github.com/rust-lang/rust/issues/42301, `rustdoc` pulls in the `html_root_url` from the rand crate, which points to GitHub Pages in rand 0.9. I don't think there is anything we can do at the moment to fix this. The next release or `rand` will remove the `html_root_url` (https://github.com/rust-random/rand/commit/b72dac1f53b0574222b430fdb8332e8b1a050d30) and hopefully rustdoc will produce the proper links to `docs.rs`.

PR #182, opened 2026-01-29
Hyperbolic overlap

mthran:
This PR implements the machinery necessary for simulating hard polygon interactions in hyperbolic space. This includes periodic boundary conditions which respect parallel transport of oriented bodies.


hoomd-geometry/src/hyperbolic_overlap.rs
```diff
@@ -0,0 +1,588 @@
...
+        // - (nu.sinh()) * (ep.cosh()) * (theta.cos()))
+        // pt[0]
+        // + (-(nu.cosh()) * ep.sinh() * theta.sin() * phi.cos()
+        // - ep.sinh() * theta.cos() * phi.sin()
+        // - nu.sinh() * ep.cosh() * theta.sin())
+        // pt[1]
+        // + ((nu.sinh()) * (ep.sinh()) * (phi.cos()) + (nu.cosh()) * (ep.cosh())) * pt[2],
+        // ]);
+        // Hyperbolic::from_minkowski_coordinates(translated, point.skirt())
+        // };
```
joaander:
```suggestion
```
Commented code confuses future developers.

hoomd-geometry/src/shape/hyperbolic_convex_polytope.rs
```diff
@@ -0,0 +1,588 @@
...
+/// A faceted hyperbolic solid defined by the convex hull of its vertices.
+#[derive(Clone, Debug, PartialEq)]
+pub struct HyperbolicConvexPolytope<const N: usize> {
+    /// The vertices of the shape
+    vertices: Vec<Hyperbolic<N>>,
+    /// The radius of the bounding sphere of the shape in the Hyperbolic metric.
+    bounding_radius: f64,
+}
+
+impl<const N: usize> HyperbolicConvexPolytope<N> {
```
joaander, line 67: `HyperbolicConvexPolytope` should be in `src/shape/hyperbolic_convex_polytope.rs`.

hoomd-geometry/src/lib.rs
```diff
@@ -117,6 +117,7 @@ use thiserror::Error;
 mod convex;
 pub use convex::Convex;
 
+pub mod hyperbolic_overlap;
```
joaander:
```suggestion
```
There should be no `hyperbolic_overlap` public module. If the code currently in this file does not all migrate to `shape/hyperbolic_convex_polytope`, then it can remain as a private module.

hoomd-geometry/src/shape/hyperbolic_convex_polytope.rs
```diff
@@ -0,0 +1,588 @@
...
+//! Implement Overlap Check for Hyperbolic Surfaces
+
+use crate::BoundingSphereRadius;
+use hoomd_manifold::{Hyperbolic, Minkowski};
+use hoomd_utility::valid::PositiveReal;
+use hoomd_vector::{Angle, Metric};
+use robust::{Coord, orient2d};
+use std::f64::consts::PI;
+
+/// A faceted hyperbolic solid defined by the convex hull of its vertices.
```
joaander, line 13: Add examples.

hoomd-geometry/src/hyperbolic_overlap.rs
```diff
@@ -0,0 +1,588 @@
...
+    /// Get the vertices of the shape.
+    #[inline]
+    #[must_use]
+    pub fn vertices(&self) -> &[Hyperbolic<N>] {
+        &self.vertices
+    }
+    /// Compute the bounding radius from a set of vertices.
+    #[inline]
+    #[must_use]
+    pub fn bounding_radius(vertices: &[Hyperbolic<N>]) -> f64 {
```
joaander:
```suggestion
    fn bounding_radius(vertices: &[Hyperbolic<N>]) -> f64 {
```
This is an internal method and should not be exposed publicly.

hoomd-geometry/src/shape/hyperbolic_convex_polytope.rs
```diff
@@ -0,0 +1,588 @@
...
+            .expect("hard coded constant should be positive")
+    }
+}
+
+/// A two-dimensional hyperbolic convex polytope.
+pub type HyperbolicConvexPolygon = HyperbolicConvexPolytope<3>;
+
+impl HyperbolicConvexPolytope<3> {
+    /// Create a regular *n*-gon with *n* vertices and a given circumradius in
+    /// hyperbolic space.
```
joaander, line 98: Add example.

hoomd-geometry/src/shape/hyperbolic_convex_polytope.rs
```diff
@@ -0,0 +1,588 @@
...
+                    Hyperbolic::<3>::from_polar_coordinates(circumradius, theta, skirt)
+                })
+                .collect::<Vec<_>>(),
+            bounding_radius: circumradius,
+        }
+    }
+    /// Calculate the distance between the center of a `HyperbolicConvexPolytope<3>`
+    /// centered at the origin and the edge at angle phi. The calculation works by
+    /// computing the intersection of the the hyperboloid with a plane passing
+    /// through the origin and the two adjacent vertices of the polygon.
```
joaander, line 115: Add example.

hoomd-geometry/src/hyperbolic_overlap.rs
```diff
@@ -0,0 +1,588 @@
...
+}
+
+/// Test whether two shapes overlap.
+pub trait SeparatingPlanes<S, M, R> {
+    /// Test whether the set of points in one shape intersects with the set of another.
+    /// Method works by iterating through adjacent vertices of `self`, constructing the
+    /// hyperplane which pass through the two vertices, and checking if all the
+    /// points of `other` are on the opposite side. If a separating plane exists, then
+    /// the two shapes do not overlap and the method returns `false`.
+    fn intersects_at(&self, x_i: &M, r_i: &R, x_j: &M, r_j: &R) -> bool;
```
joaander: I think it is confusing to offer the `intersects_at` method with a different signature in two different public traits. I see two options to fix this.
1) If these methods are truly specific to hyperbolic polytopes, then simply move them into the inherent implementation of `HyperbolicConvexPolygon`.
2) If you plan to implement related overlap checks for other hyperbolic shapes and see a need to write generic code using those implementations, we could consider splitting the current `IntersectsAt` trait. For example, there could be an `IntersectsAtGlobal` trait for `intersects_at_global` whose function signature is compatible with hyperbolic space. That would leave `intersects_at` and `approximate_separation_distance` in the `IntersectsAt` (local) trait which assume more properties about the space.

joaander (changes requested):
Thanks! I have some comments on the `hyperbolic_overlap` module that need to be addressed before this can be merged.

joaander (approved):
Thanks! I was able to remove the need for the `P` in `SitePairEnergy<S, P>` by removing the unnecessarily strict type bounds on its implementation on `HardShape`.

PR #183, opened 2026-01-30
Fix/ellipse performance

janbridley:
I swapped my root-finding code for ellipses to a root-counting algorithm instead. I also added a large number of tests, with a set of random but manually verified test cases. The first batch of these is in `test_ellipse_overlap_known_cases`, which has cases of random ellipse pairs (random axes, angles, and displacements) with distances down to ~1e-5 or so. `test_ellipse_overlap_likely_cases` has even smaller displacements with fixed ellipse axes and small angular displacements.

> NOTE: this is a breaking change, as prior to this PR the convention used by `Ellipse::intersects_at` was exactly opposite `Convex::intersects_at`. This is now fixed, and both conventions agree. This does not affect cases where the semi-major axes of both ellipses are equivalent (e.g. {a0, a1}={b0, b1}) but it DOES affect all other cases.

> NOTE: When we implement this for (hyper)ellipsoids, we can follow the same template as this approach. However, we should be able to analyze the (quartic) discriminant for real roots which will be quite efficient.

## performance before this PR (copied from #130)
```
├─ ellipsoid                        │               │               │               │         │
│  ├─ ***fast_2d         614.1 ns   │ 1.26 µs       │ 1.155 µs      │ 1.111 µs      │ 100     │ 400
│  │                  1.628 Mitem/s │ 793.5 Kitem/s │ 865.1 Kitem/s │ 900 Kitem/s   │         
   │
│  ├─ xenocollide_2d  36.37 ns      │ 51.02 ns      │ 42.56 ns      │ 42.62 ns      │ 100     │ 12800
│  │                  27.48 Mitem/s │ 19.59 Mitem/s │ 23.49 Mitem/s │ 23.45 Mitem/s │         │
│  ╰─ xenocollide_3d  111.5 ns      │ 166.2 ns      │ 132.4 ns      │ 133.8 ns      │ 100     │ 3200
│                     8.963 Mitem/s │ 6.014 Mitem/s │ 7.552 Mitem/s │ 7.469 Mitem/s │         │
```

## performance after this PR
```
Timer precision: 41 ns
overlaps              fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ ellipsoid                        │               │               │               │         │
   ├─ ***fast_2d      17.41 ns      │ 22.94 ns      │ 19.69 ns      │ 19.53 ns      │ 100     │ 25600
   │                  57.41 Mitem/s │ 43.57 Mitem/s │ 50.77 Mitem/s │ 51.17 Mitem/s │         │
   ├─ xenocollide_2d  28.48 ns      │ 37.27 ns      │ 31.9 ns       │ 32.09 ns      │ 100     │ 25600
   │                  35.1 Mitem/s  │ 26.82 Mitem/s │ 31.34 Mitem/s │ 31.15 Mitem/s │         │
   ╰─ xenocollide_3d  94.72 ns      │ 137.6 ns      │ 107 ns        │ 108.2 ns      │ 100     │ 6400
                      10.55 Mitem/s │ 7.262 Mitem/s │ 9.337 Mitem/s │ 9.234 Mitem/s │
```


joaander (approved):
Thanks!

The tutorial is now 30% faster with the `HardShape<Ellipse>` fast path.

PR #259, opened 2026-03-25
Implement Graham scan for 2D convex hulls

janbridley:
## Description

This PR adds a two dimensional convex hull algorithm based on the Graham scan, which sorts points relative to a known anchor before pruning vertices inside the hull. The upside is an `n log n` runtime, fewer calls to the numerically problematic `orient2d` predicate than gift wrapping, and a simpler implementation than quickhull.

## Design Decision (TODO)

Because the hull code requires a fixed dimension, it cannot go directly into `with_vertices`. I propose we rename the previous code to `with_vertices_unchecked` (which would result in panics if the input is empty, but this seems OK to me), and have the correctly validated code be `with_vertices`. I will write a 3d hull at some point soon, which could provide the N=3 variant of that method. @joaander , let me know what you think when you're back from break.

### predicates

I use a [modified version of the standard determinant-based orientation predicate](https://observablehq.com/@mourner/non-robust-arithmetic-as-art), which uses the coordinates of a simplex formed by an edge+point in global coordinates, rather than the simpler local coordinates.

I do not implement the fully robust predicate, although we could use the implementation Michelle uses for Voronoi -- that being said, in 2D all but the most extreme degenerate cases will be hulled correctly even without robust predicates.

The following is a plot of the output of the predicate associated with this PR, highlighting the accuracy improvements compared to the standard predicate. Note the plots show squares of side length $2^{-42}$.

| Naive Predicate | Improved Predicate |
| ------------- | ------------- |
| naive_predicate  | symmetric_predicate  |

## Motivation and context

This allows us to validate 2d convex polygon inputs, and implement separating planes for hard polygons

## How has this been tested?

New tests have been added.

## Checklist:

- [ ] I have summarized these changes in `CHANGELOG.md` following the established format.

### DIFF
--- a/hoomd-geometry/src/hull.rs
+++ b/hoomd-geometry/src/hull.rs
@@ -308,13 +308,13 @@ mod tests {
             let mut vertices1: Vec<Cartesian<2>> = (0..30)
                 .map(|_| Cartesian::from([rng.random::<f64>(), rng.random::<f64>()]))
                 .collect();
-            assert!(hull_2d_grahamscan(&mut vertices1));
+            hull_2d_grahamscan(&mut vertices1).unwrap();
 
             let mut rng = StdRng::seed_from_u64(123);
             let mut vertices2: Vec<Cartesian<2>> = (0..30)
                 .map(|_| Cartesian::from([rng.random::<f64>(), rng.random::<f64>()]))
                 .collect();
-            assert!(hull_2d_grahamscan(&mut vertices2));
+            hull_2d_grahamscan(&mut vertices2).unwrap();
 
             assert_eq!(vertices1.len(), vertices2.len());
         }
--- a/hoomd-geometry/src/hull.rs
+++ b/hoomd-geometry/src/hull.rs
@@ -174,20 +174,35 @@ mod tests {
         hull_2d_grahamscan(&mut vertices).unwrap();
         assert_eq!(vertices.len(), 4);
 
-        let hull = [[0.0, 0.0].into(), [1.0, 0.0].into(), [1.0, 1.0].into(), [0.0, 1.0].into()];
+        let hull = [
+            [0.0, 0.0].into(),
+            [1.0, 0.0].into(),
+            [1.0, 1.0].into(),
+            [0.0, 1.0].into(),
+        ];
         itertools::assert_equal(&vertices, &hull);
     }
 
     #[rstest]
     fn test_square_corners_big() {
-        let mut vertices: Vec<Cartesian<2>> = vec![[101.0, 101.0], [101.0, 100.0], [100.0, 101.0], [100.0, 100.0]]
-            .into_iter()
-            .map(Cartesian::from)
-            .collect();
+        let mut vertices: Vec<Cartesian<2>> = vec![
+            [101.0, 101.0],
+            [101.0, 100.0],
+            [100.0, 101.0],
+            [100.0, 100.0],
+        ]
+        .into_iter()
+        .map(Cartesian::from)
+        .collect();
         hull_2d_grahamscan(&mut vertices).unwrap();
         assert_eq!(vertices.len(), 4);
 
-        let hull = [[100.0, 100.0].into(), [101.0, 100.0].into(), [101.0, 101.0].into(), [100.0, 101.0].into()];
+        let hull = [
+            [100.0, 100.0].into(),
+            [101.0, 100.0].into(),
+            [101.0, 101.0].into(),
+            [100.0, 101.0].into(),
+        ];
         itertools::assert_equal(&vertices, &hull);
     }
 
@@ -209,7 +224,12 @@ mod tests {
         hull_2d_grahamscan(&mut vertices).unwrap();
         // Hull should have 4 corners (interior edge points excluded)
         assert_eq!(vertices.len(), 4);
-        let hull = [[0.0, 0.0].into(), [1.0, 0.0].into(), [1.0, 1.0].into(), [0.0, 1.0].into()];
+        let hull = [
+            [0.0, 0.0].into(),
+            [1.0, 0.0].into(),
+            [1.0, 1.0].into(),
+            [0.0, 1.0].into(),
+        ];
         itertools::assert_equal(&vertices, &hull);
     }
 
@@ -236,7 +256,12 @@ mod tests {
         hull_2d_grahamscan(&mut vertices).unwrap();
         assert_eq!(vertices.len(), 4);
 
-        let hull = [[0.0, 0.0].into(), [1.0, 0.0].into(), [1.0, 1.0].into(), [0.0, 1.0].into()];
+        let hull = [
+            [0.0, 0.0].into(),
+            [1.0, 0.0].into(),
+            [1.0, 1.0].into(),
+            [0.0, 1.0].into(),
+        ];
         itertools::assert_equal(&vertices, &hull);
     }
 
@@ -350,7 +375,12 @@ mod tests {
         hull_2d_grahamscan(&mut vertices).unwrap();
         assert!(vertices.len() >= 3);
 
-        let hull = [[0.0, 0.0].into(), [1.0, 0.0].into(), [1.0, 1.0].into(), [0.0, 1.0].into()];
+        let hull = [
+            [0.0, 0.0].into(),
+            [1.0, 0.0].into(),
+            [1.0, 1.0].into(),
+            [0.0, 1.0].into(),
+        ];
         itertools::assert_equal(&vertices, &hull);
     }
 
@@ -443,7 +473,12 @@ mod tests {
         // Should only keep furthest point in each direction
         assert!(vertices.len() >= 3);
 
-        let hull = [[0.0, 0.0].into(), [1.0, 0.0].into(), [3.0, 3.0].into(), [0.0, 1.0].into()];
+        let hull = [
+            [0.0, 0.0].into(),
+            [1.0, 0.0].into(),
+            [3.0, 3.0].into(),
+            [0.0, 1.0].into(),
+        ];
         itertools::assert_equal(&vertices, &hull);
     }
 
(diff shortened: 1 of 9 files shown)


hoomd-geometry/src/shape/convex_polytope.rs
```diff
@@ -166,6 +195,9 @@ impl<const N: usize> ConvexPolytope<N> {
         if vertices.is_empty() {
             return Err(Error::DegeneratePolytope);
         }
```
joaander:
```suggestion
```
Redundant with the added check.

hoomd-geometry/src/shape/convex_polytope.rs
```diff
@@ -166,6 +195,9 @@ impl<const N: usize> ConvexPolytope<N> {
...
+        if vertices.len() < (N + 1) {
+            return Err(Error::DegeneratePolytope);
+        }
```
joaander, line 184: What happened to the convex hull on N==2? As written, the user can use with_vertices to construct a 2D polytope with out-of-order or non-hull vertices. The resulting polytope would have undefined behavior when used with separating planes (#260).

When we add this back in, we also need to note in the `with_vertices` documentation that non-hull vertices may be removed and that 2D vertices are reordered counter-clockwise.

hoomd-geometry/src/shape/convex_polytope.rs
```diff
@@ -108,6 +108,35 @@ pub type ConvexPolygon = ConvexPolytope<2>;
...
 impl ConvexPolytope<2> {
+    /// Create a 2D convex polygon with the given vertices.
+    ///
+    /// The vertices are reduced to their convex hull using a Graham scan.
+    ///
+    /// # Errors
+    ///
+    /// Returns `[Error::DegeneratePolytope]` when the hull has fewer than 3 vertices.
+    #[inline]
+    pub fn with_vertices_hull<I>(vertices: I) -> Result<ConvexPolytope<2>, Error>
```
joaander: `with_vertices_hull` makes it sounds like the vertices must already be the hull.

How about:
```suggestion
    pub fn convex_hull_from_vertices<I>(vertices: I) -> Result<ConvexPolytope<2>, Error>
```

hoomd-geometry/src/hull.rs
```diff
@@ -0,0 +1,491 @@
...
+///
+/// # Returns
+///
+/// `()` if the hull was computed, [`Error`] if a hull could not be generated.
+///
+/// # Errors
+///
+/// Returns [`Error`] if the input vertices do not form a convex body with 3 or more points.
+#[inline]
+pub fn hull_2d_grahamscan(vertices: &mut Vec<Cartesian<2>>) -> Result<(), Error> {
```
joaander:
```suggestion
pub fn construct_2d_convex_hull(vertices: mut Vec<Cartesian<2>>) -> Result<Vec<Cartesian<2>>, Error> {
```
The name of the algorithm is not important to the function of the method. The mutable borrow is also confusing. The intent of the method is better indicated when it takes ownership of the vec and returns a modified Vec.

hoomd-geometry/src/lib.rs
```diff
@@ -445,14 +445,22 @@ pub trait MapPoint<P> {
     fn map_point(&self, point: P, other: &Self) -> Result<P, Error>;
 }
 
+/// TODO
+mod hull;
+pub use hull::hull_2d_grahamscan;
```
joaander: I understand that you want to make this method public for use outside the context of `ConvexPolytope`, but it doesn't belong at the root of the crate. Conceptually, a convex hull is intertwined with the idea of a `ConvexPolygon`, so the method should be an associated method of that type.

hoomd-geometry/src/lib.rs
```diff
@@ -445,14 +445,22 @@ pub trait MapPoint<P> {
...
 pub enum Error {
     /// Polytopes require at least one vertex.
-    #[error("a ConvexPolytope must have at least one vertex")]
+    #[error("a ConvexPolytope must have at least (N+1) vertices")]
     DegeneratePolytope,
 
+    /// Convex Polytopes must be convex
+    #[error("a ConvexPolytope must be convex")]
+    PolytopeNotConvex,
+
```
joaander: `PolytopeNotConvex` is not used anywhere.
```suggestion
```

hoomd-geometry/src/shape/convex_polytope.rs
```diff
@@ -108,6 +108,35 @@ pub type ConvexPolygon = ConvexPolytope<2>;
...
+    where
+        I: IntoIterator<Item = Cartesian<2>>,
+    {
+        let mut vertices: Vec<Cartesian<2>> = vertices.into_iter().collect();
+
+        if vertices.is_empty() {
+            return Err(Error::DegeneratePolytope);
+        }
```
joaander:
```suggestion
```
Redundant.

hoomd-geometry/src/shape/convex_polytope.rs
```diff
@@ -108,6 +108,35 @@ pub type ConvexPolygon = ConvexPolytope<2>;
 pub type ConvexPolyhedron = ConvexPolytope<3>;
 
 impl ConvexPolytope<2> {
+    /// Create a 2D convex polygon with the given vertices.
+    ///
+    /// The vertices are reduced to their convex hull using a Graham scan.
+    ///
+    /// # Errors
+    ///
+    /// Returns `[Error::DegeneratePolytope]` when the hull has fewer than 3 vertices.
```
joaander: Example needed.

joaander (changes requested):
Thanks!

I added more explicit checks to the tests to ensure that the correct hull is produced (and in the correct order), not just the correct number of vertices.

See the line-specific suggestions for API improvements.

hoomd-geometry/src/shape/convex_polytope.rs
joaander: I see, I missed this note:
> Because the hull code requires a fixed dimension, it cannot go directly into with_vertices. I propose we rename the previous code to with_vertices_unchecked (which would result in panics if the input is empty, but this seems OK to me), and have the correctly validated code be with_vertices. I will write a 3d hull at some point soon, which could provide the N=3 variant of that method. @joaander , let me know what you think when you're back from break.

FYI: edits to the description do not appear in the timeline and are therefore less visible than a comment.

We can discuss the API options offline.

janbridley:
Initial (messy) port of ConvexPolytope to ArrayVec shows measurable performance improvement, provided one sets the number of vertices as compile time. If you just use the default (32 for 2d, 64 for 3d) it's 50-80% slower than the previous code.

```diff

 ├─ polytopes                        │               │               │               │           │ 
 │  ├─ dipyramid_3d                  │               │               │               │           │
-│  │  ├─ 3+2          69.9 ns       │ 128.4 ns      │ 93.99 ns      │ 93.29 ns      │ 100     -│ 6400 vec
+│  │  ├─ 5            50.18 ns      │ 132.2 ns      │ 73.94 ns      │ 74.15 ns      │ 100     │ 6400 arrayvec
-│  │  ├─ 4+2          80.96 ns      │ 147.3 ns      │ 107 ns        │ 106.9 ns      │ 100     │ 6400 vec
+│  │  ├─ 6            54.09 ns      │ 123.7 ns      │ 83.06 ns      │ 83.3 ns       │ 100     │ 6400 arrayvec
-│  │  ├─ 8+2          119.4 ns      │ 239.1 ns      │ 156.4 ns      │ 156.7 ns      │ 100     │ 3200 vec
+│  │  ├─ 10           90.56 ns      │ 219.4 ns      │ 130.2 ns      │ 131.5 ns      │ 100     │ 3200 arrayvec
-│  │  ├─ 16+2         189.7 ns      │ 371.9 ns      │ 252.1 ns      │ 250.1 ns      │ 100     │ 3200 vec
+│  │  ├─ 18           181.6 ns      │ 362.6 ns      │ 242.9 ns      │ 241.2 ns      │ 100     │ 3200 arrayvec
-│  │  ╰─ 50+2         546.4 ns      │ 1.546 µs      │ 707.9 ns      │ 743.1 ns      │ 100     │ 800 vec
+│  │  ╰─ 52           640 ns        │ 1.593 µs      │ 796.2 ns      │ 825.7 ns      │ 100     │ 800 arrayvec

 │  ╰─ polygon_2d                    │               │               │
-│     ├─ 3            99.2 ns       │ 112.8 ns      │ 101.1 ns      │ 101.8 ns      │ 100     │ 6400 vec
+│     ├─ 3            57.52 ns      │ 109.6 ns      │ 58.18 ns      │ 59.04 ns      │ 100     │ 12800 arrayvec
-│     ├─ 4            129.8 ns      │ 141.5 ns      │ 133.7 ns      │ 134.8 ns      │ 100     │ 3200 vec
+│     ├─ 4            75.76 ns      │ 82.27 ns      │ 77.71 ns      │ 78.01 ns      │ 100     │ 6400 arrayvec
-│     ├─ 8            204 ns        │ 283.4 ns      │ 207.9 ns      │ 209.7 ns      │ 100     │ 3200 vec
+│     ├─ 8            137.6 ns      │ 262.6 ns      │ 141.5 ns      │ 143.4 ns      │ 100     │ 3200 arrayvec
-│     ├─ 16           364.2 ns      │ 408.4 ns      │ 374.5 ns      │ 376.5 ns      │ 100     │ 1600 vec
+│     ├─ 16           330.2 ns      │ 364.1 ns      │ 335.5 ns      │ 336.5 ns      │ 100     │ 1600 arrayvec
-│     ╰─ 50           1.259 µs      │ 1.707 µs      │ 1.353 µs      │ 1.364 µs      │ 100     │ 400 vec
+│     ╰─ 50           1.228 µs      │ 1.562 µs      │ 1.332 µs      │ 1.338 µs      │ 100     │ 400 arrayvec
```

joaander:
> If you just use the default (32 for 2d, 64 for 3d) it's 50-80% slower than the previous code.

I don't see any differences in the research-relevant benchmarks. I consistently get 118.6 steps/s (on `cheme-hodges`) for the octahedron benchmark on `trunk`, this branch with `ConvexPolyhedron` and this branch with `ConvexPolytope<3, 6>`. The hexagon benchmark consistently runs at 392 steps/s.

I think what you are seeing in `cargo bench` is some time related to passing the owned `ArrayVec` from one stack frame to another in:
```
            .bench_local_values(|((p0, p1), (t, r))| black_box(collide3d(&p0, &p1, &t, &r)));
```

In `overlap.rs`, we should create only 1 pair of shapes in each of the `dipyramid_3d` and `polygon_2d` scopes. The `collide?d(&p0, &p1` calls will then capture a reference to those and we never need to trigger a copy of the full `ArrayVec`. To make the overlap checks more realistic, `create_offset_3d(&mut rng)` should choose a random position in a ring with r between 0.5 * 2 * circumradius and 2 * circumradius.

I don't see a need for the shapes themselves to be random.

Hopefully these changes can give us better performance data from `cargo bench`.

joaander (approved):

PR #284, opened 2026-05-04
Implement a modified Gram-Schmidt orthonormalization algorithm

janbridley:
## Description

- Useful as a fast but less numerically stable approach to things like QR. Can also be used in testing QR, and for [LLL](https://en.wikipedia.org/wiki/Lenstra–Lenstra–Lovász_lattice_basis_reduction_algorithm) which I need for a personal project

## Motivation and context

Resolves #???

## Checklist:


tfteague:
Code-wise, everything looks good to me to be merged in to box. Should we add a disclaimer about usage (e.g. when to prefer using QR)? Also, I wonder, how far do we want to go in terms of implementing & supporting LA algorithms as part of hoomd-rs vs. having the users supply their own code for these operations?

janbridley:
> Code-wise, everything looks good to me to be merged in to box. Should we add a disclaimer about usage (e.g. when to prefer using QR)? Also, I wonder, how far do we want to go in terms of implementing & supporting LA algorithms as part of hoomd-rs vs. having the users supply their own code for these operations?

I'll add a note re: using the other QR code. IN terms of supporting LA more generally, my goal is to implement what we have a need/clear use case for but I can definitely see a more minimal set being valuable

PR #287, opened 2026-05-05
implement Ord (and Eq) for PositiveReal

janbridley:
## Description

PositiveReal types avoid the ambiguity present in floating point comparisons, allowing for a correct implementation of `Ord` rather than just `PartialOrd`. This PR is a minimal implementation, using `total_cmp` for the implementation. Note we could probably improve performance by casting to u64 and doing comparisons there but it seems like overkill.

## Motivation and context

This allows for ::max and ::min reductions (and heaps, etc) based on positive real types

## Checklist:

### DIFF
--- a/doc/src/release-notes.md
+++ b/doc/src/release-notes.md
@@ -4,6 +4,8 @@
 
 *Added:*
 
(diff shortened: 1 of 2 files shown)


joaander (approved):
Thanks!

PR #288, opened 2026-05-05
Implement BoundingSphereRadius for Simplex3

janbridley:
## Description

Implement a minimal centered bounding sphere radius for the Simplex3 type.

@joaander -- is this correct in that we should return the origin-centered bounding sphere? Or is a local bounding sphere radius correct? The current code seems to match ConvexSurfaceMesh2d but I figured I would ask.

> NOTE: if we do not care to handle NaN (which ConvexSurfaceMesh2d and ConvexPolytope do not), we can simplify the code. Performance is fine either way, but I want to make sure computing the radius for pre-overlap filtering works well.

## Checklist:


joaander (approved):
Thanks!

PR #293, opened 2026-05-19
Spherical Harmonics in Rust

janbridley:
## Description

Simple and efficient evaluation of spherical harmonics in rust. While a crate for this does exist, it appears to be abandonware and panics for `l > 10`.

## Motivation and context

Evaluation of the Steinhardt order parameters requires spherical harmonics, and the Glotzerlab's current codebase ([fsph](https://fsph.readthedocs.io/en/latest/))for spherical harmonics is not structured to efficiently calculate things like this. With some compile-time generics and some of the new findings from the [`sphericart`](https://sphericart.readthedocs.io/en/latest/maths.html) paper in JCP, I have an approach better-suited to our standard workflow. It provides all positive m values for a compile-time value of L, and can efficiently evaluate both individual and multiple points. A doc example shows a simple Steinhardt calculation, although the real implementation of that should go in hoomd-order or similar.

### Example implementation of q6

```rust
/// Implement the Steinhardt order parameter q6.
fn q6(bonds: &[Cartesian<3>]) -> f64 {
    let mut accum = [Complex64::ZERO; 7];
    let y6 = SphericalHarmonic::<6>::new();

    for &bond in bonds {
        let projected_onto_sphere = bond / bond.norm();
        let qlmi = y6.eval(projected_onto_sphere.coordinates);
        for m in 0..7 { accum[m] += qlmi[m]; }
    }

    // We multiply the `m>0` components by two to account for `-m` contributions.
    let sum_sq = accum[0].norm_sqr()
        + 2.0 * accum[1..].iter().map(Complex64::norm_sqr).sum::<f64>();
    let n = bonds.len() as f64;

    (4.0 * PI / 13.0 * sum_sq).sqrt() / n
}
```

## How has this been tested?

Tests against [SPHRS](https://github.com/argmin-rs/sphrs) for l<=10 compare the actual values, and I have an integration-ish test that tests the [completeness relation](https://en.wikipedia.org/wiki/Spherical_harmonics#Completeness_Relation) to measure accumulated numerical error. Offline/mixed-language tests have shown my results are similar accuracy to fsph and sphericcart, but better performance when only a few l values are required.

## Checklist:


hoomd-order/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+        Self::new()
+    }
+}
+
+/// Complex spherical harmonics `Y_L^m` for a single degree L.
+///
+/// Index with `[m]` to access `Y_L^m` for m = 0..=L.
+/// The m = 0 term is always purely real.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct HarmonicOutput<const L: usize> {
```
mthran: Is there a particular reason for separating the m=0 from the other m values?

hoomd-order/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+    /// The length of the container, equal to `L + 1`.
+    #[inline]
+    #[must_use]
+    pub const fn len(&self) -> usize {
+        L + 1
+    }
+    /// Check if the container is empty. This will always be false.
+    #[inline]
+    #[must_use]
+    pub const fn is_empty(&self) -> bool {
```
mthran, line 355: When would this be used?

mthran (commented):
Very nice! Other than a couple clarifying questions, this API is consistent with how I would hope to use spherical harmonics.

hoomd-order/src/lib.rs
janbridley: Ideally the full output could be stored as a `[Complex64; L + 1]`, but generating an array of that size is quite complex on stable rust. This is a workaround, although m=0 is structurally different from m>0 so it seemed a reasonable compromise. I could be convinced otherwise though!

hoomd-order/src/lib.rs
janbridley: Basically never, as there is no way to construct the object manually and `len` is always > 0. However, I got a clippy lint warning when I didn't implement it -- see below. Maybe we disable the warning for this case? I agree it is meaningless.
image

joaander: Clippy also issues lints that suggest using `is_empty` instead of `len() == 0` on containers. Even if it seems superfluous, we should keep `is_empty` because it is idiomatic rust and less effort than applying clippy ignores in many places.

janbridley:
The last major question (and the failing lint) is a readme -- Josh, I see this crate as useful outside `hoomd-rs` but the maintenance burden of it being fully standalone seems undesirable. Is it worth having the README (and therefore crates.io page) be more or less independent and just release the code with `hoomd-rs`, or should we just make this a core part of the HOOMD code? Per some offline discussions with Michelle, the actual order parameters using this code will go in a crate like `hoomd-order`

rsph/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+            normalized_recurrence_seed,
+            z_coeff,
+            rxy_coeff,
+        }
+    }
+
+    /// Evaluate `Y_L^m` for m = 0..=L at a point on the unit sphere.
+    #[must_use]
+    #[inline]
+    pub fn eval(&self, point: [f64; 3]) -> HarmonicOutput<L> {
```
joaander: What if `point` is not on the unit sphere?

rsph/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+    /// assert_abs_diff_eq!(full[0].conj(), full[2 * out.l()])
+    /// ```
+    #[inline]
+    pub fn iter(&self) -> impl Iterator<Item = Complex64> + '_ {
+        std::iter::once(self.m0).chain(self.mp.iter().copied())
+    }
+    /// Compute `Y_l^{-m} = (-1)^m · conj(Y_l^m)`.
+    #[inline]
+    #[must_use]
+    pub fn neg_m(&self, m: usize) -> Complex64 {
```
joaander: I think `neg_m` is an awkward approach because the API for accessing positive and negative m values is not symmetric. I have two ideas for a symmetric API:

1. Remove the `Index` implementation. Access all `m` values via: `output.positive(m)` or `output.negative(m)`. This is clean and simple, but still breaks symmetry with `iter()` so that might also have to be removed.

2. To keep `Index` and `iter`, implement them for newtypes. `HarmonicOutput` would not allow direct access to the output, but `Positive(&output)` and `Negative(&output)` would implement `Index` and `iter`: e.g. `Positive(&output)[1]`. You could implement methods that return the newtype: `output.positive()[0]`, `output.negative().iter()`.

I'm open to other suggestions.

joaander (commented):
Thanks! This is clean and simple. We discussed offline regarding what crate makes the most sense to hold this code.

You could make use of `katex` math in the documentation to make it more clear and explicit what this code calculates.

rsph/src/lib.rs
janbridley: Fair question -- we could use `Unit` here, which is probably the cleanest fix?

joaander: What about spherical coordinates?

rsph/src/lib.rs
janbridley: Discussed offline: remove neg_m, keep iterator and document that it iterates over positive m.

joaander:
The spherical harmonic code is complete. I moved this PR back to draft state as it now serves as the creation of the `hoomd-order` crate. I will merge this branch after additional PRs that add basic functionality to `hoomd-order`.

PR #305, opened 2026-06-01
Implement a four-dimensional cross-product analog

janbridley:
## Motivation and context

This approach generalizes the concept of the cross product to arbitrary dimension, although I have only implemented it for N=4 as most algorithms for this scale exponentially with N. This code is planned to be used in a four-dimensional xenocollide implementation, although I tried to make the implementation as domain-agnostic as possible. It is possible we could rewrite the current Cross trait in terms of this more general feature, but we run into N-1 generic issues.

## How has this been tested?

New tests validate invariants including orthogonality and the volume of the spanned vectors.

## Checklist:


janbridley:
@josephburkhart , I know you've spent some time thinking about geometric algebra. Does this definition and way of calculating this quantity make sense to you?

Edit: note that it is not simply the wedge product, and calculating it as the hodge dual of the wedge product is fairly inefficient compared the the determinant-based formula

joaander (approved):
Thanks!

PR #306, opened 2026-06-02
Xenocollide in 4 Dimensions

janbridley:
## Description

Implement a four-dimensional variant of the Xenocollide collision detection algorithm.

## Motivation and context

Resolves #???

## Checklist:


hoomd-geometry/src/shape/convex_polytope.rs
```diff
@@ -377,6 +377,15 @@ impl<const N: usize, const MAX_VERTICES: usize> ConvexPolytope<N, MAX_VERTICES>
     }
 }
 
+impl<const MAX_VERTICES: usize> Volume for ConvexPolytope<2, MAX_VERTICES> {
+    #[inline]
+    fn volume(&self) -> f64 {
```
joaander: I would prefer that users explicitly make the conversion to `ConvexSurfaceMesh2d` to compute volume. That gives them a chance to catch the error from try_from and respond instead of forcing a panic.

janbridley:
Ok, so the code prior to [f5aba4b](https://github.com/glotzerlab/hoomd-rs/pull/306/commits/f5aba4b4dc82d6bc79e5daf360fa4d241c6aee28) is correct as far as I can tell, but it left a small ambiguity in how we selected the portal after narrowing. Previously, we did a two-step check to identify which facet the origin exits through and then identify whether or not the ray actually penetrated that facet. In doing so, a very small percentage of configurations could cycle, where we would select the incorrect exiting face (for example, if the origin ray passes through two facets in quick succession) and rapidly swapped back and forth between those configurations, hitting the iteration cap and safely exiting with a collision detection. Longer period cycles were also possible, but I don't have a good intuition for what exactly caused those. After looking back over the math and the lower dimensional cases, I believe the current implementation is now exact (up to floating point weirdness) and can no longer cycle, just as in the 2D and 3D cases. Note that we did not observe this behavior in those lower dimensions because the counary cross product provides a simple closed form for the ray exit test.

As a nice benefit, I think this new approach('s code, the mathematics are obviously transfterrable) generalizes trivially to higher dimensions.

PR #346, opened 2026-06-30
Implement `DoubleVersor`, a representation of rotations in four dimensions

janbridley:
## Description

Rotations in four dimensions have a nice representation as a pair of quaternions, forming denoting the left- and right-isoclinic components of the rotation. Intuitively, we can think about (a limited subset of) SO(3) as a rotation *in* the x-y plane. To reach every possible rotation, we must allow that plane to vary, giving us the axis-angle representation. SO(4) is similar, where we rotation *about* the x-y and z-w planes. Like in SO(3), we must allow both of these planes to change in order to allow for arbitrary rotations - but crucially, they rotate independently. Below 4D, this uncoupling is not possible, but in 4 and higher dimensions you always have at least a few independent hyperplanes to rotate about.

## Motivation and context

https://github.com/user-attachments/assets/94b03eac-10c7-496d-99c6-376ed44f613e

## How has this been tested?

New tests, plus a small Mathematica snippet for the metric code

## Checklist:


janbridley:
Pinging you Michelle as I think you have the most knowledge of manifold algebra. Let me know if you have any questions!

hoomd-mc/src/rotate.rs
mthran, line 1: Nit-picky: I think it may be helpful to specify the units of `maximum_rotation` for `DoubleVersor`

hoomd-vector/src/lib.rs
```diff
@@ -163,6 +180,8 @@
 //! - Angles are uniformly sampled from the half-open interval `[0, 2π)`
 //! - Versors are uniformly sampled from the surface of the `3-Sphere`, which doubly
 //!   covers `SO(3)`, the manifold of rotations in three dimensions.
+//! - Double versors are sampled as independent pairs of versors, which uniformly
+//!   covers `SO(4)`, the manifold of rotations in four dimensions.
```
mthran, line 184: I think it might be helpful to say a little more here (though this may be redundant, I could believe that anyone interested in `DoubleVersor` already knows what SO(4) is!). Maybe a quick line like:

Rotations in 4-dimensional space are classified into two types: ``simple" and ``double'' rotations. Simple rotations are those which leave an entire two-dimensional plane fixed (analogously to how 3-dimensional rotations leave a one-dimensional axis fixed). Double rotations can be described as two (nonzero and unequal) rotations of two orthogonal planes. Geometrically, the SO(4) manifold is isomorphic to (SU(2) \times SU(2) )/ \mathbb{Z}_2.

hoomd-vector/src/doubleversor.rs
```diff
@@ -0,0 +1,621 @@
...
+use rand::{Rng, RngExt};
+use rand_distr::{Distribution, StandardUniform};
+use serde::{Deserialize, Serialize};
+
+use crate::{Cartesian, Metric, Quaternion, Rotate, Rotation, RotationMatrix, Versor};
+
+/// A pair of [`Versor`]s that represent a 4D rotation.
+///
+/// Every rotation in SO(4) factors into a *left-isoclinic* and a *right-isoclinic*
+/// rotation via the Van Elfrinkhof decomposition, each of which can be described as an
```
mthran, line 22: Again, I am not sure if this is too niche, but if you came across a good reference for this, it might be helpful to include a hyperlink here!

hoomd-vector/src/doubleversor.rs
```diff
@@ -0,0 +1,621 @@
...
+        // (* Contrain our solutions to the manifold *)
+        // constraints = {
+        //   a1^2 + b1^2 + c1^2 + d1^2 == 1, a2^2 + b2^2 + c2^2 + d2^2 == 1,
+        //   p1^2 + q1^2 + r1^2 + s1^2 == 1, p2^2 + q2^2 + r2^2 + s2^2 == 1
+        // };
+
+        // (* Validate *)
+        // Print["Norm[A-B, \"Frobenius\"]^2 == 8 - 8*(qL1.qL2)*(qR1.qR2)"];
+        // isEquivalent = FullSimplify[explicitFrobeniusSquared == algebraicForm, constraints];
+        // Print["Result: ", isEquivalent];
```
mthran, line 184: Thank you for this comment, I would have been extremely confused otherwise. This is cool!!

mthran (commented):
My comments are probably not very helpful, I mostly picked at documentation. Looks good to me overall!

hoomd-vector/src/quaternion.rs
```diff
@@ -600,16 +608,20 @@ impl Versor {
...
+    /// $`\lVert \vec{v} \rVert = \sin\phi`$, so this equals the standard equation
+    /// $`\arccos(\mathbf{q}_0 \cdot \mathbf{q}_1)`$ exactly.
     #[inline]
     #[must_use]
     pub fn arc_distance(&self, other: &Self) -> f64 {
-        self.dot_as_cartesian(other).acos()
+        let relative_quat = *self.inverted().combine(other).get();
+        f64::atan2(relative_quat.vector.norm(), relative_quat.scalar)
     }
+
```
joaander: Why is this change necessary if it is exactly equivalent?

hoomd-vector/src/quaternion.rs
```diff
@@ -304,6 +304,14 @@ impl Quaternion {
...
         Versor(self / self.norm())
     }
+
+    /// Embed this quaternion as a point in R^4.
+    #[inline]
+    #[must_use]
+    pub fn embed_in_cartesian_4(&self) -> Cartesian<4> {
+        let [x, y, z] = self.vector.coordinates;
+        [self.scalar, x, y, z].into()
+    }
```
joaander:
```suggestion
```
This is only called in one place, you can move this code there. Also, we choose explicitly to not offer this transformation as a public API because the order of the components is ambiguous.

hoomd-vector/src/lib.rs
```diff
@@ -163,6 +180,13 @@
 //! - Angles are uniformly sampled from the half-open interval `[0, 2π)`
 //! - Versors are uniformly sampled from the surface of the `3-Sphere`, which doubly
 //!   covers `SO(3)`, the manifold of rotations in three dimensions.
+//! - Double versors are sampled as independent pairs of versors, which uniformly
+//!   covers `SO(4)`, the manifold of rotations in four dimensions. Four-dimensional
+//!   rotations are either *simple*, leaving an entire two-dimensional plane fixed
+//!   (analogously to how three-dimensional rotations leave a one-dimensional axis
+//!   fixed), or *double*, rotating two orthogonal planes by independent angles. This is
+//!   fundementally different from lower-dimensional rotation groups, which do not admit
+//!   these additional rotational behaviors.
```
joaander:
```suggestion
//!   covers `SO(4)`, the manifold of rotations in four dimensions.
```
This is good information, but is not necessary when describing how `DoubleVersor` is sampled via `StandardUniform`.

hoomd-vector/src/lib.rs
```diff
@@ -148,6 +148,23 @@
...
+//!     PI,
+//! ));
+//! assert_relative_eq!(v.rotate(&a), [-2.0, 1.0, 0.0, 0.0].into());
+//! # Ok(())
+//! # }
+//! ```
+//!
 //! Convert to a [`RotationMatrix`] when you need to rotate many vectors by the same
 //! rotation. [`RotationMatrix::rotate`] is typically several times faster than
 //! [`Versor::rotate`].
```
joaander: This paragraph refers to `Versor`. It should go *above* the description of `DoubleVersor`.

hoomd-vector/src/doubleversor.rs
```diff
@@ -0,0 +1,622 @@
...
+    /// let v = DoubleVersor::default();
+    /// ```
+    #[inline]
+    fn default() -> Self {
+        Self {
+            l: Versor::default(),
+            r: Versor::default(),
+        }
+    }
+}
```
joaander:
```suggestion
#[derive(Clone, Copy, Debug, Default, PartialEq, RelativeEq, Serialize, Deserialize)]
pub struct DoubleVersor {
    /// The left-isoclinic part of the rotation.
    l: Versor,
    /// The right-isoclinic part of the rotation.
    r: Versor,
}
```

hoomd-vector/src/doubleversor.rs
```diff
@@ -0,0 +1,622 @@
...
+
+impl From<(Versor, Versor)> for DoubleVersor {
+    #[inline]
+    fn from(value: (Versor, Versor)) -> Self {
+        Self {
+            l: value.0,
+            r: value.1,
+        }
+    }
+}
```
joaander:
```suggestion
```
Why invent new representations of something that is already well defined? Make the fields of `DoubleVersor` public instead so that callers can use struct initialization syntax.

You can then also remove the `left_isoclinic`, `right_isoclinic`, `from_left_isoclinic`, and `from_right_isoclinic` methods which are now redundant.

hoomd-vector/src/doubleversor.rs
```diff
@@ -0,0 +1,622 @@
...
+/// \vec{a}' = \mathbf{q}_l \, \vec{a} \, \mathbf{q}_r
+/// ```
+/// As normal [`Versor`]s do with SO(3), the pair $`(\mathbf{q}_l, \mathbf{q}_r)`$ forms
+/// a double cover of SO(4).
+#[derive(Clone, Copy, Debug, PartialEq, RelativeEq, Serialize, Deserialize)]
+pub struct DoubleVersor {
+    /// The left-isoclinic part of the rotation.
+    l: Versor,
+    /// The right-isoclinic part of the rotation.
+    r: Versor,
```
joaander:
```suggestion
    pub q_l: Versor,
    /// The right-isoclinic part of the rotation.
    pub q_r: Versor,
```
Name fields consistent with the documentation.

hoomd-vector/src/doubleversor.rs
```diff
@@ -0,0 +1,622 @@
...
+/// of freedom. [`DoubleVersor`] stores one unit quaternion for each factor,
+/// $`\mathbf{q_l}`$ and $`\mathbf{q_r}`$. The two rotate a [`Cartesian<4>`] vector
+/// $`\vec{a}`$ (viewed as a quaternion) with a double-sided product that mirrors the
+/// three-dimensional [`Versor::rotate`]:
+///
+/// ```math
+/// \vec{a}' = \mathbf{q}_l \, \vec{a} \, \mathbf{q}_r
+/// ```
+/// As normal [`Versor`]s do with SO(3), the pair $`(\mathbf{q}_l, \mathbf{q}_r)`$ forms
+/// a double cover of SO(4).
```
joaander: Add a rustdoc example that initializes a `DoubleVersor` with struct initialization syntax.

hoomd-vector/src/doubleversor.rs
```diff
@@ -0,0 +1,622 @@
...
+        }
+    }
+
+    #[inline]
+    #[must_use]
+    /// Normalize the double versor.
+    ///
+    /// Nominally, all [`DoubleVersor`] instances have unit components. Due to limited
+    /// floating point precision, this assumption may not hold after repeated
+    /// operations. Normalize double versors when needed to correct this issue.
```
joaander: Add a rustdoc example that normalizes a `DoubleVersor`.

hoomd-vector/src/doubleversor.rs
```diff
@@ -0,0 +1,622 @@
...
+    /// Explicitly, this is the straight-line (chordal) distance between two
+    /// rotations $`\vec{u}`$ and $`\vec{v}`$. It is NOT the intrinsic distance
+    /// along the manifold, which is defined by the [`Metric`] implementation.
+    ///
+    /// ```math
+    /// d_{SO(4)}(\vec{u}, \vec{v}) = \sqrt{8 \left(1 - (\vec{u}_l \cdot \vec{v}_l)\,(\vec{u}_r \cdot \vec{v}_r)\right)}
+    /// ```
+    ///
+    /// This is equivalent to the matrix form $`\lVert R_u - R_v \rVert_F`$, but does not require
+    /// forming the matrix representation of each rotation.
```
joaander:
```suggestion
    /// This is equivalent to the matrix form $`\lVert R_u - R_v \rVert_F`$
    /// where $` R_u `$ is, ....
```
There is no need to discuss implementation details in the user-facing documentation.

hoomd-vector/src/doubleversor.rs
```diff
@@ -0,0 +1,622 @@
...
+    /// # Example
+    ///
+    /// ```
+    /// use approxim::assert_relative_eq;
+    /// use hoomd_vector::{DoubleVersor, Metric, Versor};
+    /// use std::f64::consts::PI;
+    ///
+    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// let identity = DoubleVersor::default();
+    /// // A rotation by PI/2 in both invariant planes.
```
joaander:
```suggestion
```
No need to restate what the code already says.

hoomd-vector/src/doubleversor.rs
```diff
@@ -0,0 +1,622 @@
...
+    /// let v = DoubleVersor::from_left_isoclinic(Versor::from_axis_angle(
+    ///     [1.0, 0.0, 0.0].try_into()?,
+    ///     PI,
+    /// ));
+    ///
+    /// assert_relative_eq!(v.chordal_distance(&identity), 8.0_f64.sqrt());
+    /// // The intrinsic distance along the manifold is longer than the chord.
```
joaander:
```suggestion
```
No need to restate what the code already says.

hoomd-vector/src/doubleversor.rs
```diff
@@ -0,0 +1,622 @@
...
+        let (a, [b, c, d]) = (q_l.scalar, q_l.vector.coordinates);
+        let (p, [q, r, s]) = (q_r.scalar, q_r.vector.coordinates);
+
+        // Construct the left-isoclinic matrix L(Q_L)
+        let l_mat = [[a, -b, -c, -d], [b, a, -d, c], [c, d, a, -b], [d, -c, b, a]];
+
+        // Construct the right-isoclinic matrix R(Q_R)
+        let r_mat = [[p, -q, -r, -s], [q, p, s, -r], [r, -s, p, q], [s, r, -q, p]];
+
+        // Combine the left and right isoclinic parts as L@R
```
joaander:
```suggestion
        let l_mat = [[a, -b, -c, -d], [b, a, -d, c], [c, d, a, -b], [d, -c, b, a]];
        let r_mat = [[p, -q, -r, -s], [q, p, s, -r], [r, -s, p, q], [s, r, -q, p]];

```
These comments only restate the code.

hoomd-vector/src/doubleversor.rs
```diff
@@ -0,0 +1,622 @@
...
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> DoubleVersor {
+        (rng.random(), rng.random()).into()
+    }
+}
+
+impl Rotation for DoubleVersor {
+    /// Combine two rotations.
+    ///
+    /// The resulting versor is obtained by left and right quaternion
+    /// multiplications. As with [`Versor`], `self.combine(other)` first applies
```
joaander:
```suggestion
    /// multiplications. `self.combine(other)` first applies
```
Not just "as with Versor", but it is clearly stated in the contract of `combine` as defined in the `Rotation` trait.

hoomd-vector/src/doubleversor.rs
```diff
@@ -0,0 +1,622 @@
...
+    /// let left = DoubleVersor::from_left_isoclinic(Versor::from_axis_angle(
+    ///     [1.0, 0.0, 0.0].try_into()?,
+    ///     PI / 2.0,
+    /// ));
+    /// let right = DoubleVersor::from_right_isoclinic(Versor::from_axis_angle(
+    ///     [0.0, 1.0, 0.0].try_into()?,
+    ///     PI / 3.0,
+    /// ));
+    ///
+    /// // Combining applies `right` first, then `left`.
```
joaander:
```suggestion
```

hoomd-vector/src/doubleversor.rs
```diff
@@ -0,0 +1,622 @@
...
+    /// lengths, folding the sum into $`[0, \pi]`$ to account for the double
+    /// cover of the manifold:
+    /// * $`\alpha_l = \arccos(\vec{u}_L \cdot \vec{v}_L)`$
+    /// * $`\alpha_r = \arccos(\vec{u}_R \cdot \vec{v}_R)`$
+    /// * $`\theta_1 = \min(\alpha_l + \alpha_r, 2\pi - (\alpha_l + \alpha_r))`$
+    /// * $`\theta_2 = |\alpha_l - \alpha_r|`$
+    ///
+    /// This is equivalent to the scaled matrix logarithm form $`||\log(R_u^T R_v)||_F`$,
+    /// measuring the shortest path along the curved manifold.
+    fn distance(&self, other: &Self) -> f64 {
```
joaander:
```suggestion

    /// The intrinsic distance between two [`DoubleVersor`]s.
    ///
    /// Explicitly, the metric for two points $`\vec{u}`$ and $`\vec{v}`$ on the
    /// manifold SO(4):
    ///
    /// ```math
    /// d_{SO(4)}(\vec{u}, \vec{v}) = \sqrt{2\theta_1^2 + 2\theta_2^2}
    /// ```
    ///
    /// Where the principal planar angles $`\theta_1`$ and $`\theta_2`$ of the
    /// relative rotation are derived from the left and right quaternion arc
    /// lengths, folding the sum into $`[0, \pi]`$ to account for the double
    /// cover of the manifold:
    /// * $`\alpha_l = \arccos(\vec{u}_L \cdot \vec{v}_L)`$
    /// * $`\alpha_r = \arccos(\vec{u}_R \cdot \vec{v}_R)`$
    /// * $`\theta_1 = \min(\alpha_l + \alpha_r, 2\pi - (\alpha_l + \alpha_r))`$
    /// * $`\theta_2 = |\alpha_l - \alpha_r|`$
    ///
    /// This is equivalent to the scaled matrix logarithm form $`||\log(R_u^T R_v)||_F`$,
    /// measuring the shortest path along the curved manifold.
    #[inline]
    fn distance(&self, other: &Self) -> f64 {
```
Easier to read this way.

hoomd-vector/src/doubleversor.rs
```diff
@@ -0,0 +1,622 @@
...
+            let rows = RotationMatrix::from(dv).rows();
+
+            for (i, row_i) in rows.iter().enumerate() {
+                assert_relative_eq!(row_i.dot(row_i), 1.0, epsilon = 1e-14);
+                for row_j in rows.iter().skip(i + 1) {
+                    assert_relative_eq!(row_i.dot(row_j), 0.0, epsilon = 1e-14);
+                }
+            }
+        }
+    }
```
joaander: These are good self-consistency checks, but what about some basic checks that rotating a given vector by a given double versor matches an expected output (e.g. from Mathematica)?

hoomd-vector/src/cartesian.rs
```diff
@@ -816,6 +816,15 @@ impl<const N: usize> From<RotationMatrix<N>> for Matrix<N, N> {
...
 
+impl<const N: usize> From<Matrix<N, N>> for RotationMatrix<N> {
+    #[inline]
+    fn from(value: Matrix<N, N>) -> Self {
+        Self {
+            rows: value.rows.map(Cartesian::from),
+        }
+    }
+}
+
```
joaander:
```suggestion
```
unsafe

hoomd-mc/src/rotate/doubleversor.rs
```diff
@@ -0,0 +1,196 @@
...
+    /// The standard deviation of the normal distribution of quaternions around the identity.
+    std_dev: f64,
+}
+
+impl From<f64> for DoubleVersorDisplacement {
+    #[inline]
+    fn from(value: f64) -> Self {
+        Self { std_dev: value }
+    }
+}
```
joaander:
```suggestion
pub(crate) struct DoubleVersorDisplacement {
    /// The standard deviation of the normal distribution of quaternions around the identity.
    pub(crate) std_dev: f64,
}
```
Use struct initialization or a `new` method to construct `DoubleVersorDisplacement`. `From` should only be implemented when the conversion is *value preserving* and *obvious*.

Also, does this need to be `pub(crate)`? It is used in the same module.

hoomd-mc/src/rotate/doubleversor.rs
```diff
@@ -0,0 +1,196 @@
...
+    use assert2::check;
+    use hoomd_microstate::property::OrientedPoint;
+    use hoomd_vector::{Cartesian, DoubleVersor, Metric};
+    use rand::{SeedableRng, rngs::StdRng};
+    use rstest::*;
+
+    /// Number of trial moves to test.
+    const N: usize = 262_144;
+
+    /// Expected geodesic distance for small a:E[χ₆] = sqrt2*Γ(7/2)/Γ(3) = 15sqrt(2π)/16
```
joaander: Even if it is not rendered, LaTeX would be more readable than this.

hoomd-mc/src/rotate/doubleversor.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+impl Adjust for Rotate<DoubleVersor> {
+    /// Change the maximum trial move size by the given scale factor.
+    #[inline]
+    fn adjust(&mut self, factor: PositiveReal) {
+        self.maximum_rotation *= factor;
+
+        if self.maximum_rotation.get() > PI / 2.0 {
+            self.maximum_rotation = (PI / 2.0)
```
joaander: I suppose this is an internal usage case where `positive_real!` would work well.

joaander (changes requested):
Thanks! I have many suggestions to improve this, mostly around the public-facing API.

hoomd-vector/src/quaternion.rs
janbridley: The acos formulation is numerically unstable near 0 distance. I will note that in the docs, or this could be a `// comment` so it does not show up in the public docs?

joaander: When I search for this method, the first hit is a stack overflow post that uses `atan2`. I'm not sure where we get the idea that `acos` is "standard" and needs to be compared to in the first place.

hoomd-vector/src/doubleversor.rs
janbridley: I think this is non-obvious, as the rotation is half the versor's rotation angle. I will expand on the docs to make some of these nuances more clear

hoomd-vector/src/quaternion.rs
janbridley: To me acos of a dot product is the standard way to find an angle between vectors but that may be my own bias. I'll clean up the doc.

hoomd-vector/src/doubleversor.rs
janbridley: I added a deeper discussion to the primary struct docs, so this should be self-documenting now. Thank you!

janbridley:
That should be everything! Please let me know if any more changes are required.

PR #359, opened 2026-07-10
Implement a subset of signac

joaander:
## Description

Implement a minimal subset of signac's functionality in Rust. Organize the code so that more functionality can be added over time as needed.

## Motivation and context

It would be convenient if the entire simulation workflow from initialization to analysis could be performed in Rust. Currently, users need Python to initialize a signac workspace for use with row and hoomd-rs.

This pull requests implements a minimal set of functionality so that the initialization and job state point parsing can be handled in Rust. Specifically, it implements:
* Create job
* Open job from directory
* Open job from state point

Compatibility is maintained so that any job written by the Rust implementation is valid in the Python implementation and any job written by the Python implementation is valid in Rust. The latter may not be convenient, but it should be possible. In highly heterogeneous workspaces, users will need to parse state points as `serde_json::Value`.

The Rust implementation will follow ideas currently under consideration for signac 3.0, such as immutable state points (https://github.com/glotzerlab/signac/issues/983), and workspaces that include *only* job directories.

## Scope

The minimal signac methods will be placed in the `hoomd-workspace` crate. If we choose to expand the scope, we can later move them to their own crate, independent of hoomd-rs. To keep the scope minimal, `hoomd-workspace` will have no notion of a project directory. The job methods will operate assuming that the current working directory is the project directory.

## How has this been tested?

* Unit tests with reference job ids computed by the Python implementation.
* Manually check that projects created in Rust are usable in Python.

## Checklist:

- [X] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-rs/blob/trunk/CONTRIBUTING.md).
- [X] I agree with the terms of the [**hoomd-rs Contributor Agreement**](https://github.com/glotzerlab/hoomd-rs/blob/trunk/ContributorAgreement.md).
- [X] My name is on the list of contributors (`doc/src/credits.md`) in the pull request source branch.
- [x] I have summarized these changes in `release-notes.md` following the established format.


josephburkhart:
I forgot that Github has a special review interface - I made some suggested changes, all docs-related, and put them all in a commit. Take a look and see what you think. Other than those suggestions, I think this is good to go!

josephburkhart (approved):

PR #420, opened 2026-09-11
Document hoomd-order design. Implement psi_k and q_L order parameters.

joaander:
## Description

Building on work @janbridley started in #293, this PR implements the k-atic and Steinhardt order parameters. It also sketches the overall design for future order parameter methods.

## Motivation and context

Allow users to compute order parameters entirely within the Rust toolchain. For example, users could use order parameters in metadynamics simulations. See `ARCHITECTURE.md` for a complete motivation.

## How has this been tested?

New unit tests.

## Checklist:

- [X] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-rs/blob/trunk/CONTRIBUTING.md).
- [X] I agree with the terms of the [**hoomd-rs Contributor Agreement**](https://github.com/glotzerlab/hoomd-rs/blob/trunk/ContributorAgreement.md).
- [X] My name is on the list of contributors (`doc/src/credits.md`) in the pull request source branch.

### DIFF
--- a/.codebook.toml
+++ b/.codebook.toml
@@ -1,4 +1,5 @@
 words = [
+    "Steinhardt",
     "abelian",
     "binaryen",
     "bindgen",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4000,6 +4000,7 @@ dependencies = [
  "anyhow",
  "approxim",
  "divan",
+ "hoomd-geometry",
  "hoomd-microstate",
  "hoomd-spatial",
  "hoomd-vector",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3999,6 +3999,8 @@ version = "1.3.0"
 dependencies = [
  "approxim",
  "divan",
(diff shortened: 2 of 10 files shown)


joaander:
I still need to complete the overview docs and address some remaining TODO items, but the overall API structure is ready to review now. Start with `ARCHITECTURE.md`.

@janbridley take a look and let me know what you think. The design for the new "neighbor query" types was heavily influenced by ideas that you and @DomFijan came up with for freud. To make it work, I needed to keep the concept of site tags present in some data structures (see `average_over_neighbors`) so that site identity could be tracked across multiple invocations. Instead of representing per-site data as an array in site index order, it is a `HashMap` that maps from the site tag to the order parameter. This scheme will work naturally when users e.g. compute order parameters only on sites with type A. The resulting sparse `HashMap` will retain the identity of the sites for use in further downstream processing (e.g. clustering) or visualization.

PR #421, opened 2026-09-11
Add macro to generate positive_real values in const contexts

janbridley:
## Description

 Construct a `PositiveReal` from a const-evaluatable value.

To do this, I added a `from_f64` value (mirrors char::from_u32, except we return result instead of option). This can be const, where `try_from` cannot be (because it's a trait impl). Currently its hidden from the docs for clarity, but the impl is infallible and could be a first-class method if we want it to be.

 The value is validated during compilation: it must be finite and greater than zero or the invocation fails to compile. Values that are not known at compile time must be constructed with `PositiveReal::try_from` instead.

```rust
 use hoomd_utility::valid::PositiveReal;

 const HALF: PositiveReal = hoomd_utility::positive_real!(0.5);
 const SUM: PositiveReal = hoomd_utility::positive_real!(1.0 + 1.0);

 assert_eq!(HALF.get(), 0.5);
 assert_eq!(SUM.get(), 2.0);

```

 This does not compile:

 ```rust
 use hoomd_utility::valid::PositiveReal;

 const BAD: PositiveReal = hoomd_utility::positive_real!(-1.0);
 ```

## Motivation and context

Resolves #???

## Checklist:

- [x] I have summarized these changes in `release-notes.md` following the established format.

### DIFF
--- a/hoomd-utility/src/valid/positive_real.rs
+++ b/hoomd-utility/src/valid/positive_real.rs
@@ -16,6 +16,8 @@ use super::Error;
 ///
 /// # Example
 ///
+/// Construct from a value known at runtime with [`PositiveReal::try_from`]:
+///
 /// ```
 /// use hoomd_utility::valid::PositiveReal;
 ///
@@ -24,10 +26,39 @@ use super::Error;
 /// # Ok(())
 /// # }
 /// ```
+///
+/// Construct from a value known at compile time with [`crate::positive_real!`]:
+///
+/// ```
+/// use hoomd_utility::valid::PositiveReal;
+///
+/// const HALF: PositiveReal = hoomd_utility::positive_real!(0.5);
+/// const SUM: PositiveReal = hoomd_utility::positive_real!(1.0 + 1.0);
+///
+/// assert_eq!(HALF.get(), 0.5);
+/// assert_eq!(SUM.get(), 2.0);
+/// ```
 #[derive(Clone, Copy, Debug, PartialEq, Serialize, Deserialize)]
 pub struct PositiveReal(f64);
 
 impl PositiveReal {
+    /// Construct from a value, checking that it is finite and greater than 0.
+    ///
+    /// This is a const-callable equivalent of [`PositiveReal::try_from`] used
+    /// by the [`crate::positive_real!`] macro, which cannot build the value
+    /// directly because the field is private.
+    #[doc(hidden)]
+    #[inline]
+    pub const fn from_f64(v: f64) -> Result<Self, Error> {
+        if !v.is_finite() {
+            Err(Error::NotFinite(v))
+        } else if v <= 0.0 {
+            Err(Error::NotPositive(v))
+        } else {
+            Ok(Self(v))
+        }
+    }
+
     /// Access the value.
     ///
     /// # Example
@@ -43,7 +74,7 @@ impl PositiveReal {
     /// # }
     #[must_use]
     #[inline]
-    pub fn get(&self) -> f64 {
+    pub const fn get(&self) -> f64 {
         self.0
     }
 }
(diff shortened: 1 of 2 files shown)


hoomd-utility/src/valid/positive_real.rs
```diff
@@ -24,10 +26,39 @@ use super::Error;
...
 #[derive(Clone, Copy, Debug, PartialEq, Serialize, Deserialize)]
 pub struct PositiveReal(f64);
 
 impl PositiveReal {
+    /// Construct from a value, checking that it is finite and greater than 0.
+    ///
+    /// This is a const-callable equivalent of [`PositiveReal::try_from`] used
+    /// by the [`crate::positive_real!`] macro, which cannot build the value
+    /// directly because the field is private.
+    #[doc(hidden)]
```
joaander:
```suggestion
    /// This is a const-callable equivalent of [`PositiveReal::try_from`].
```

I don't see a need to hide the docs. There is also no need to explain internal implementation details to the user.

hoomd-utility/src/valid/positive_real.rs
```diff
@@ -98,16 +129,49 @@ impl TryFrom<f64> for PositiveReal {
...
+///
+/// This does not compile:
+///
+/// ```compile_fail
+/// use hoomd_utility::valid::PositiveReal;
+///
+/// const BAD: PositiveReal = hoomd_utility::positive_real!(-1.0);
+/// ```
+#[macro_export]
+macro_rules! positive_real {
```
joaander, line 159: Why does the macro export to the crate root? Is is possible to name it `hoomd_utility::valid::positive_real` to allow room for future growth in `hoomd_utility`?

joaander (changes requested):
Thanks, this is a great addition!

Should we encourage the use of `positive_real!` in the tutorials and examples? `positive_real!(value)` is longer than `value.try_into()?` and it requires a `use` at the top, but there are many ancillary benefits to compile time checks.

janbridley:
> Thanks, this is a great addition!
> 
> Should we encourage the use of `positive_real!` in the tutorials and examples? `positive_real!(value)` is longer than `value.try_into()?` and it requires a `use` at the top, but there are many ancillary benefits to compile time checks.

I think it's worth switching, I personally feel like I just `expect` most positive real conversions as it's usually easy to guarantee correctness. If we can prove that at compile time, all the better

hoomd-utility/src/valid/positive_real.rs
janbridley: Agreed - this is an oversight on my part. I'll try to update that

janbridley: Ah, looks like I did have a reason: https://github.com/rust-lang/rust/issues/39412 , https://github.com/rust-lang/rust/issues/145140 . We can work around this with some doc[hidden] shenanigans, or put the macro in hoomd_derive and import from there (at the cost of putting syn into the dependency tree of hood-utility. I would lean toward the doc hack, but I agree its unsatisfying.

joaander: I suppose we can leave it where it is so that we don't have to use arcane workarounds.

joaander (approved):
Thanks!

I went ahead and changed the main examples to use `positive_real` where possible. There are cases where `try_into` is still necessary: such as computing the box volume based on the shapes volume --- the shape volume methods are not `const`. There are a few cases where the box length is computed completely from constants and built-in math methods that *could* be const, but I left them as `try_into`.

I didn't bother modifying the many doc code examples. I don't think the benefit to that is worth the cost.

janbridley:
> Thanks!
> 
> I went ahead and changed the main examples to use `positive_real` where possible. There are cases where `try_into` is still necessary: such as computing the box volume based on the shapes volume --- the shape volume methods are not `const`. There are a few cases where the box length is computed completely from constants and built-in math methods that _could_ be const, but I left them as `try_into`.
> 
> I didn't bother modifying the many doc code examples. I don't think the benefit to that is worth the cost.

I can send an LLM to switch things out where possible if that's helpful? Otherwise I agree its not worth the effort to find and replace.

joaander:
The goal of the rustdoc examples is to provide a minimal code that works. In this case `.try_into` is a valid choice.

The goal of the tutorials and complete examples is to demonstrate best practices, which is why it makes sense to use `positive_real!` there.

PR #426, opened 2026-09-18
Implement Convex hull algorithms in 2 and 3 dimensions; add basic ConvexSurfaceMesh3d type

janbridley:
## Description

Our existing 2d hull code works well, but was not built to generalize the interface in higher dimensions. I have added a new `trait ConvexHull`, which is implemented for `Cartesian<2|3>`, and returns both vertices and facets of the hull. The `Facet<N>` type is an N-1-dimensional simplex, stored as usize vertex indices into an auxiliary array. This gets around const generic math issues and gives us a minimal storage representation, although note that I discard the edges returned from the 2d hull as ConvexSurfaceMesh2d does not need them.

- [ ] TODO: check if pre-stored edges have a performance benefit in 2d overlap benches

## Motivation and context

Hulls are basically required for accurate volume and moment of inertia calculations. There are also basically 0 good rust convex hull crates, with most being some combination of abandoned, imprecise, or massive bindings to qhull/cgal. This is a reasonably simple approach, which uses exact geometric predicates to ensure coplanar faces -- which are very common in HPMC -- are handled correctly. Both 2D and 3D use an incremental algorithm, in 2d specifically it is a Graham scan and in 3d the naive incremental hull, (roughly) as you'll find in O'Rourke's computational geometry book. This is `O(n_vertices*n_faces)`, so not asymptotically optimal but practically the fastest for the types of shapes we care about.

## How has this been tested?

Tests against a large number of high-symmetry bodies at random orientations, plus edge case and invariance tests.

## Checklist:


janbridley:
Josh - this is my best attempt at a uniform API for convex hulls, as well as a simple start on a three-dimensional explicit hull type (`ConvexSurfaceMesh3d`). Let me know your thoughts!

PR #427, opened 2026-09-18
Add four-dimensional orientation predicate with exact arithmetic

janbridley:
## Description

Implement a robust geometric predicate for hyperplane orientation in 4D.

## Motivation and context

In 2D and 3D, we use `georust::robust::orient<2|3>d` in our hull algorithm to ensure numerical consistency. These use [Shewchuk's exact predicates](https://www.cs.cmu.edu/~quake/robust.html), which are one of the fastest algorithms you can achieve but are monstrously complicated. There are papers on automatically generating these predicates, and [somewhat simpler approaches](http://www-sop.inria.fr/members/Sylvain.Pion/progs/modular/modular_eng), but these are fairly challenging to implement and validate. 

Instead, this PR implements the predicates exactly, using fixed-width integers to represent the f64 coordinates. This restricts the allowed values to scale variance `~2^72`, but in practice handles all common cases. For simplicial solids (those with vertices in general position e.g. tetrahedral facets), these checks should never trigger so the really degenerate cases are quite rare.

I tried to make things as well-typed as possible, so I implemented a basic `Dyad` type in hoomd-utilities, which is used as an intermediate between raw f64 and the [`i256::I512`](https://docs.rs/i256/latest/i256/struct.I512.html) type we use for exact comparisons. This did end up fairly complex, but there are fewer magic numbers than a comparable Shewchuk approach.

## How has this been tested?

Tested with a large number of oriented polyhedra embedded in 4d (similar to oriented polygons embedded in 3d). The tests are an absolute mess right now - they really need the DoubleVersor PR and a full hull implementation to make sense.

## Checklist:


hoomd-geometry/src/lib.rs
```diff
@@ -129,6 +129,7 @@ pub use convex::Convex;
 mod hull;
 pub use hull::{ConvexHull, Facet};
 
+pub mod orient4d;
```
joaander: I need to resolve one question before I can review the `hoomd-geometry` changes any further. This method looks like an internal implementation detail of the convex hull algorithm. Is it public intentionally? If so, why?

hoomd-utility/src/dyad.rs
```diff
@@ -0,0 +1,141 @@
...
+///
+/// let dyad = Dyad::try_from_f64(16.0).expect("the value is finite");
+/// // 16 = 2^52 * 2^-48
+/// assert_eq!(dyad.mantissa(), 1_i128 << 52);
+/// assert_eq!(dyad.exponent(), -48);
+/// ```
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub struct Dyad {
+    /// The integer significand.
+    mantissa: i128,
```
joaander: Why is the mantissa `i128`? `i64` can represent all the values in the 53-bit mantissa. I realize that you are performing higher-bit integer math with this later, but you can widen to `i128` at the point of use.

hoomd-utility/src/dyad.rs
```diff
@@ -0,0 +1,141 @@
...
+        mantissa: 0,
+        exponent: -1074,
+    };
+
+    /// Decompose a finite f64 into `mantissa · 2^exponent` exactly.
+    ///
+    /// Returns [`None`] when `value` is not finite. Signed zeros have zero mantissa.
+    #[expect(clippy::missing_panics_doc, reason = "panic is unreachable")]
+    #[inline]
+    pub fn try_from_f64(value: f64) -> Option<Self> {
```
joaander: I think you could implement the `TryFrom` trait for this type. That would be appropriate because the conversion is value preserving and obvious. 

Otherwise:
```suggestion
    pub fn from_f64(value: f64) -> Result<Self, Error> {
```
Fallible `from` inherent methods don't need the `try_` prefix. See https://doc.rust-lang.org/std/str/fn.from_utf8.html and https://doc.rust-lang.org/std/primitive.u64.html#method.from_str_radix for examples.

In either case, the return value should be a `Result`. If you need it as an `Option` later, there are methods to convert `Result` to `Option`.

hoomd-utility/src/dyad.rs
```diff
@@ -0,0 +1,141 @@
+// Copyright (c) 2024-2026 The Regents of the University of Michigan.
+// Part of hoomd-rs, released under the BSD 3-Clause License.
+
+//! An exact dyadic decomposition of floating point numbers.
+
+/// An exact dyadic rational `mantissa · 2^exponent`.
+///
+/// Every finite f64 is exactly representable in this form, and [`Dyad::try_from_f64`]
+/// finds such a decomposition with `|mantissa| ≤ 2^53 - 1` (and `mantissa == 0` only
```
joaander: This and other public methods should use LaTeX math to improve the documentation's readability.

joaander (changes requested):
Thanks!

I reviewed the `hoomd-utility` changes. We need to resolve whether `orient4d` should be public before I review the `hoomd-geometry` changes.

